PR #29, opened 2022-07-06
CRAYSAT-1408: Add variable substitution to `sat bootprep`

haasken-hpe:
## Summary and Scope

Add the ability to perform variable substitution in certain fields of
`sat bootprep` input files using Jinja2 templates. Variables can be
loaded from the following three locations in order from lowest to
highest precedence:

* In branches corresponding to versions of the HPC software recipe
  repo that have been imported into gitea (VCS) on the system. The
  version of the HPC software recipe is specified by `--recipe-version`.
* In YAML files specified by the `--vars-file` command-line option.
* Variables passed directly on the command line using the `--vars`
  option.

Various general functionality was added to support this:

* The ability to clone repositories was added to the `VCSRepo` class in
  the `apiclient.vcs` module, and that class was refactored slightly.
* A `set_val_by_path` function was added to support setting values in a
  dictionary by a dotted path.
* A `deep_update_dict` function was added to support deep dictionary
  updates or merges.
* A `sat.recipe` module was added for loading HPC Software Recipe data.
  This will likely be useful for `sat showrev` and other commands in the
  future.

Update the man page for `sat bootprep` to describe the new `--vars`,
`--vars-file`, and `--recipe-version` options. Also add missing doc for
`--bos-version` option. Add examples of using the new variable options.

Update bootprep input file schema property descriptions to indicate
which fields support rendering with Jinja template engine. Deduplicate
the branch property definition by defining it in refs and referring to
it in product- and git-based layer definitions.

## Issues and Related PRs

* Resolves [CRAYSAT-1408](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1408)

### Tested on:

  * groot
  * local macbook

### Test description:

Tested on groot-ncn-m001 by running `sat bootprep run` with various
options to specify different variable sources and ensured precedence
between the different variable sources works as documented.

Viewed rendered `sat bootprep` man page. Viewed `sat bootprep run
--help` usage information.

Generated `sat bootprep` input file schema HTML docs and viewed in
browser.

## Risks and Mitigations

No known risks.

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated


haasken-hpe:
I don't see why license-checker is failing. It's flagging `sat/cli/bootprep/vars.py`, but it's fine with `sat/recipe.py`, and yet both have the same license text:

```
(sat) C02ZH00ALVCJ: sat (CRAYSAT-1408-product-version-substitutions) $ diff -u <(head -n 23 sat/recipe.py) <(head -n 23 sat/cli/bootprep/vars.py)
(sat) C02ZH00ALVCJ: sat (CRAYSAT-1408-product-version-substitutions) $
```

Running the recommended docker command just results in duplicated license text in `vars.py`.

haasken-hpe:
> I don't see why license-checker is failing. It's flagging `sat/cli/bootprep/vars.py`, but it's fine with `sat/recipe.py`, and yet both have the same license text:
> 
> ```
> (sat) C02ZH00ALVCJ: sat (CRAYSAT-1408-product-version-substitutions) $ diff -u <(head -n 23 sat/recipe.py) <(head -n 23 sat/cli/bootprep/vars.py)
> (sat) C02ZH00ALVCJ: sat (CRAYSAT-1408-product-version-substitutions) $
> ```
> 
> Running the recommended docker command just results in duplicated license text in `vars.py`.

Oh, now it's working. I pushed a commit that added the period after the copyright notice. The checks all went away on the PR. Then they came back, and license-check was still "failed". However, now I see it is successful, so I must have just caught it at a bad time.

haasken-hpe:
Full testing description can be found in [this comment in the Jira](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1408?focusedCommentId=4907327&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4907327).

sat/recipe.py
```diff
@@ -0,0 +1,178 @@
...
+    def __eq__(self, other):
+        return self.version == other.version
+
+    def __lt__(self, other):
+        return self.version < other.version
+
+    @cached_property
+    def all_vars(self):
+        """dict: the variables defined by the products"""
+        with self.clone_content() as clone_dir:
```
jack-stanek-hpe, line 85: This would likely require more instrumentation around the Gitea API, but we could potentially use the [`/repos/{owner}/{repo}/contents/{filepath}`](https://try.gitea.io/api/swagger#/repository/repoGetContents) endpoint from Gitea to retrieve just the contents of the recipe file, instead of the whole repository. Cloning the whole repo could potentially slow things down depending on how large it gets or how long the history becomes, though depending on how small it stays this could also be a premature optimization.

sat/apiclient/vcs.py
```diff
@@ -133,12 +170,26 @@ def get_commit_hash_for_branch(self, branch):
...
-        """str: a full, git-clone-able URL to the repository"""
-        return urlunparse(('https', self.vcs_host, self.repo_path, '', '', ''))
+        Args:
+            branch (str, optional): the branch to set as HEAD in the cloned repo.
+                If omitted, defaults to the HEAD of the remote repo.
+            directory (str, optional): the directory to clone the repo to.
+                If omitted, clone to a directory matching the name of the repo.
+            single_branch (bool, optional): whether to clone just a single branch.
+        """
+        git_cmd = ['git', 'clone']
```
jack-stanek-hpe, line 183: Would it make sense to put `--depth=1` in the command options?

sat/recipe.py
```diff
@@ -0,0 +1,178 @@
...
+            HPC software recipes.
+        vcs_repo (VCSRepo): the VCS repository containing the HPC software
+            recipes
+    """
+
+    def __init__(self):
+        """Create a new HPCSoftwareRecipeCatalog."""
+        self.clone_url = urlunparse(
+            ParseResult(scheme='https', netloc=get_config_value('api_gateway.host'),
+                        path=f'{HPC_SOFTWARE_RECIPE_REPO_PATH}',
```
jack-stanek-hpe, line 124:
```suggestion
                        path=HPC_SOFTWARE_RECIPE_REPO_PATH,
```

sat/cli/bootprep/input/configuration.py
```diff
@@ -25,22 +25,48 @@
...
+    @functools.wraps(func)
+    def wrapper(self, *args, **kwargs):
+        unrendered_result = func(self, *args, **kwargs)
+
+        # If the value is `None`, it was not specified by the user.
+        if unrendered_result is None:
+            return unrendered_result
+
+        try:
+            return Template(unrendered_result).render(self.var_context.vars)
```
jack-stanek-hpe, line 62: I don't remember if you mentioned this in the testing (Jira is down at the moment), but if a variable is missing from `self.var_context.vars`, will this just replace it with the empty string? e.g. if the input template has the following somewhere:
```version: {{misspelled_product_name.version}}```
Would that just result in the following, which is semantically invalid?
```layer.product_version == ""```
If this is the case, would it make sense to pass `undefined=jinja2.StrictUndefined` to the `Template()` constructor, and handle resulting [`UndefinedError`](https://jinja.palletsprojects.com/en/3.1.x/api/#jinja2.StrictUndefined)s?

File: tests/apiclient/test_vcs.py [hunk 5]
```diff
@@ -24,27 +24,52 @@
...
-        self.assertEqual(repo.username, VCSRepo._default_username)
+    def test_repo_path(self):
+        """Test repo_path property of VCSRepo."""
+        self.assertEqual(self.repo_path, self.vcs_repo.repo_path)
+
+    def test_repo_name(self):
+        """Test repo_name property of VCSRepo."""
+        self.assertEqual(self.repo_name, self.vcs_repo.repo_name)
+
+    def test_vcs_username(self):
```
jack-stanek-hpe, line 65: w.r.t. the username-in-URL bug above, there should probably be a unit test to check that behavior is correct.

tests/cli/bootprep/input/test_configuration.py
```diff
@@ -307,6 +338,16 @@ def test_product_version_property_not_present(self):
         """Test the product_version property when version is not in the layer data"""
         self.assertEqual(LATEST_VERSION_VALUE, self.branch_layer.product_version)
 
+    def test_product_version_jinja_template(self):
+        """Test the product_version property when it uses Jinja2 templating"""
+        # Have to double the literal brackets that make up the Jinja2 variable reference
+        self.version_layer_data['product']['version'] = f'{{{{{self.product_name}.version}}}}'
```
jack-stanek-hpe, line 344: Would this be more readable?
```suggestion
        self.version_layer_data['product']['version'] = '{{' + f'{self.product_name}.version' + '}}'
```

tests/cli/bootprep/input/test_configuration.py
```diff
@@ -364,6 +406,16 @@ def test_branch_property_not_present(self):
         """Test the branch property when branch is not in the layer data"""
         self.assertIsNone(self.version_layer.branch)
 
+    def test_branch_property_jinja_template(self):
+        """Test the branch property when the branch uses Jinja2 templating"""
+        # Have to double the literal brackets that make up the Jinja2 variable reference
+        self.branch_layer_data['product']['branch'] = f'integration-{{{{{self.product_name}.version}}}}'
```
jack-stanek-hpe, line 412: Same question as above, is this more readable?
```suggestion
        self.version_layer_data['product']['branch'] = 'integration-{{' + f'{self.product_name}.branch' + '}}'
```

tests/test_recipe.py
```diff
@@ -0,0 +1,127 @@
...
+            self.assertIn(recipe_version, recipe_catalog.recipes)
+            recipe = recipe_catalog.recipes[recipe_version]
+            self.assertEqual(recipe_version, recipe.version)
+            self.assertEqual(f'cray/hpc-software-recipe/{recipe_version}',
+                             recipe.vcs_branch)
+            self.assertEqual(self.mock_vcs_repo, recipe.vcs_repo)
+
+    def test_recipes_property_vcs_error(self):
+        """Test the recipes property when there is an error accessing VCS."""
+        type(self.mock_vcs_repo).remote_refs = PropertyMock(side_effect=VCSError)
```
jack-stanek-hpe, line 89: Is `type(self.mock_vcs_repo)` just `MagicMock`? If so, could this affect other tests? I might be misunderstanding why modifying the class is needed in this instance.

jack-stanek-hpe (approved):
This looks good to me, though I did have a few questions about the implementation.

sat/apiclient/vcs.py
```diff
@@ -75,38 +140,10 @@ def remote_refs(self):
...
-
-        os.chmod(cred_helper_script.name, 0o700)  # Make executable by user
-
-        env = dict(os.environ)
-        env.update(GIT_ASKPASS=cred_helper_script.name)
         try:
-            proc = subprocess.run(['git', 'ls-remote', vcs_full_url],
-                                  stdout=subprocess.PIPE, stderr=subprocess.PIPE,
-                                  env=env, check=True)
+            proc = self.run_authenticated_git_cmd(['git', 'ls-remote', self.user_clone_url])
```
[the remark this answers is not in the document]
haasken-hpe: I agree with you that we wouldn't have a way to handle authentication if given a different username than what is specified in the `vcs-user-credentials` K8s secret.

As a result, I've added code to `VCSRepo.__init__` that will check whether a username is specified in the `netloc` of the `clone_url`, and if so, it will discard the username and log a warning message to indicate it is being ignored. I think it would also make sense to do this validation as an earlier step when processing the CFS configuration layers, so we could be really clear about which layer in the input file has the problem, but this will be easier to do as a part of the refactor in [CRAYSAT-1278](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1278). I added a comment there as a reminder.

sat/util.py
```diff
@@ -460,6 +460,92 @@ def get_val_by_path(dict_val, dotted_path, default_value=None):
...
+                    'last_name': 'Carey'
+                }
+                'genre': 'Game Show'
+            }
+        }
+
+    Returns:
+        None. The dictionary `dict_val` is modified in place.
+    """
+    def _set_val_by_path_helper(current_dict, split_path, val):
```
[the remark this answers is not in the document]
haasken-hpe: I converted this tail-recursive function into an iterative approach, and I think it's clearer now too.

tests/apiclient/test_vcs.py
haasken-hpe: Done, thanks!

tests/cli/bootprep/input/test_configuration.py
haasken-hpe: Yes, I think that helps distinguish between the brackets that are part of the Jinja template and the ones that are used by the fstring

tests/test_recipe.py
haasken-hpe: I'm not exactly sure how this works, but I believe the way I'm doing it is consistent with the [PropertyMock docs](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.PropertyMock).

Example of it not affecting other `MagicMock` objects:

```
>>> from unittest.mock import MagicMock, PropertyMock
>>> m1 = MagicMock()
>>> p = PropertyMock(return_value=3)
>>> type(m1).foo = p
>>> m1.foo
3
>>> m2 = MagicMock()
>>> m2.foo
<MagicMock name='mock.foo' id='4496337120'>
```

sat/recipe.py
haasken-hpe: Agree that this would be a nice future optimization. I do have some future hopes for which clone functionality would be useful. Specifically, it would be nice to have some automation around helping the admin create and merge the `integration` branches of these repos, perhaps as another mode of `sat bootprep`.

sat/apiclient/vcs.py
haasken-hpe: I think I'll add it as an argument to this function, and I'll use it in `HPCSoftwareRecipe.clone_content` since we really do only need the tip of that branch in that case.

sat/cli/bootprep/input/configuration.py
haasken-hpe: Summarizing the discussion from scrum, if we try to access a key in an undefined variable, e.g. `version: {{misspelled_product_name.version}}`, we will get an error. If we access just a variable that doesn't exist, it will be replaced by an empty string. I think that's probably okay. The dotted variable paths are the more common use case right now.

haasken-hpe: I will be giving this more thought and testing as I go on an implement other variables though. I added a note to CRAYSAT-1411 to make sure I consider this then when I'm implementing additional non-dict variables.

PR #34, opened 2022-07-20
CRAYSAT-1409: Allow `sat bootprep` to find image/recipe from product

haasken-hpe:
### Allow `sat bootprep` to find image/recipe from product

Update the `sat bootprep` input file schema to allow the user to specify
images which are created from images or recipes provided by a product.
Update the `sat bootprep` code to query the `cray-product-catalog` K8s
ConfigMap to find the image/recipe provided by a product.

Note that this does not handle products which deliver multiple recipes
or images yet. If an image or recipe is requested from a product, and it
delivers more than one image or recipe, an error will be returned to the
user.

Due to the bootprep schema update, validation unit tests had to be
updated to find errors at their new indentation levels since the
top-level error for an invalid image will be of the form "Not valid
under any of the given schemas", and the more specific error messages
will be indented an additional level.

Refactor the `render_template` function which was only used by the
`sat.cli.bootprep.input.configuration` module to a shared location and
rename to `jinja_rendered` to sound more like a decorator. This
decorator is now used in CFS configurations (and layers) and in images
from the input instance.

Also use a `SandboxedEnvironment` when rendering Jinja2 templates to
avoid possible security vulnerabilities that could arise from rendering
untrusted Jinja2 template provided by the user. The risk of this sort of
attack using `sat bootprep` is low though given that SAT is a local
command-line interface which runs as the user invoking it.

### Add bootprep input file schema version

Add a schema version to the bootprep input file JSON schema definition
itself. Add a `schema_version` property to the `sat bootprep` input
files. Add validation to check whether the `schema_version` property
specified by input files is compatible with the current bootprep input
file schema definition.

## Issues and Related PRs

* Resolves [CRAYSAT-1409](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1409)

### Tested on:

  * groot
  * vale

### Test description:

Unit tests and pycodestyle pass.

Built docker image and using that docker image on vale, tested the
following scenarios:

* Creation of CFS configurations which specified product version
  variables.
* Creation of image from a base recipe provided by the cos product.
* Customization of an image from a base image provided by the cos
  product.
* Mocked out a product catalog entry to have no recipes/images, and
  specified the image and recipe from that product. Error resulted.
* Mocked out a product catalog entry to have multiple recipes/images,
  and specified the recipe/image from that product. Error resulted.
* Tried a malicious Jinja template of the form `"".__class__.__mro__`
  to validate that the `SandboxedEnvironment` provides some protection
  against malicious Jinja2 templates supplied by the user.

## Risks and Mitigations

None

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated


sat/cli/bootprep/input/image.py
```diff
@@ -135,29 +170,27 @@ def configuration_group_names(self):
...
-    def ims_data(self):
-        """dict: the data that defines the base IMS image or recipe"""
-        return self.image_data['ims']
-
-    @property
+    @abstractmethod
     def base_is_recipe(self):
         """bool: whether the starting point is a recipe in IMS or an image"""
-        return self.ims_data['is_recipe']
+        pass
```
jack-stanek-hpe, line 176: FWIW, the `pass` statements here and in the other abstract methods aren't necessary; the docstring is the entire function body.

sat/cli/bootprep/input/image.py
```diff
@@ -479,3 +493,124 @@ def completed_successfully(self):
...
+        except APIError as err:
+            raise ImageCreateError(err)
+
+        if len(matching_base_resources) == 0:
+            raise ImageCreateError(f'Found no matches for {self.base_description}')
+        elif len(matching_base_resources) > 1:
+            raise ImageCreateError(f'Found {len(matching_base_resources)} matches for {self.base_description}. '
+                                   f'This {self.base_resource_type} must be specified by "id".')
+
+        return matching_base_resources[0]
```
jack-stanek-hpe, line 539: This isn't a problem here, but maybe a note for once we adopt Python 3.10 -- this would be a simple candidate for using the new [structural pattern matching syntax](https://peps.python.org/pep-0636/). This could be replaced perhaps somewhat more compactly with:
```python
match len(matchign_base_resources):
    case 1:
        return matching_base_resources[0]
    case 0:
        raise ImageCreateError(f'Found no matches for {self.base_description}')
    case int(n):
        raise ImageCreateError(f'Found {n} matches for {self.base_description}. '
                               f'This {self.base_resource_type} must be specified by "id".')

```

sat/cli/bootprep/input/image.py
```diff
@@ -479,3 +493,124 @@ def completed_successfully(self):
...
+        """str: the product name"""
+        # name is a required property
+        return self.image_data['product']['name']
+
+    @property
+    @jinja_rendered
+    def product_version(self):
+        """str: the product version"""
+        # version is not required and should default to the latest version
+        return self.image_data['product'].get('version')
```
jack-stanek-hpe, line 562: This looks to me as though this will return `None` if `'version'` is missing from `image_data`, not the latest version. I may be missing a level of indirection here, though.

sat/cli/bootprep/input/image.py
```diff
@@ -479,3 +493,124 @@ def completed_successfully(self):
...
+        except APIError as err:
+            raise ImageCreateError(err)
+
+        if len(matching_base_resources) == 0:
+            raise ImageCreateError(f'Found no matches in IMS for {self.base_description} '
+                                   f'with {ims_resource_desc}')
+        elif len(matching_base_resources) > 1:
+            # This really shouldn't happen given that we are querying by unique IMS id
+            raise ImageCreateError(f'Found {len(matching_base_resources)} matches for '
+                                   f'{self.base_description} with {ims_resource_desc}')
```
jack-stanek-hpe, line 614: nit: irrespective of my comment about structural pattern matching above, it might be good for now to have a consistent style for these sorts of conditionals checking if the list has exactly 1 element. Specifically, there are instances of both checking `len(some_list) == 0` and `not some_list` in this code. FWIW I typically prefer the latter style, but since we're specifically checking the length here I see the rationale for using the former. I don't feel strongly, but I do think we should be consistent regardless.

sat/data/schema/bootprep_schema.yaml
```diff
@@ -34,6 +51,38 @@ description: >
...
+    description: >
+      The version of the bootprep input file schema used by this input file. The
+      version of the schema should be a SemVer-like version number with three
+      components, major.minor.patch. This will be populated with the correct
+      value in input files generated by 'sat bootprep generate-example'.
+
+      This version will be checked against the current schema version understood
+      by the 'sat bootprep' commmand.
+
+      Compatibility is defined as follows. Let us refer to the schemaVersion
```
jack-stanek-hpe, line 64: nit: should `schemaVersion` be camelCase here?

jack-stanek-hpe (approved):
Looks good to me. Just had a few nitpicks and suggestions for the future.

sat/cli/bootprep/input/image.py
haasken-hpe: Cool, thanks for the heads up. I wasn't aware of this new feature.

sat/cli/bootprep/input/image.py
haasken-hpe: Right, the two places this `product_version` property is used would be in the `base_description` property and in the `installed_product` cached property. The `base_description` property has logic to detect whether `self.product_version` is None. The `installed_product` cached property calls [this function in cray-product-catalog](https://github.com/Cray-HPE/cray-product-catalog/blob/a1cc1f6233616d6d45c9a3f2dfb428a8ccf02532/cray_product_catalog/query.py#L133-L170), which treats `version=None` as the latest version. I will clarify this comment to something like this:

```
# version is not required. base_description and installed_product treat None as latest
```

sat/cli/bootprep/input/image.py
haasken-hpe: I don't see a specific reason why I need to use `len(matching_base_resources) == 0` here vs. `not matching_base_resources`, so I will use the latter.

sat/data/schema/bootprep_schema.yaml
haasken-hpe: Nope, I think I originally had it as camelCase due to some confusion in my head between the camelCase properties used to define a schema using JSON Schema (i.e. properties in the JSON Schema metaschema). Then I realized this is just a new property in the bootprep input file schema, and those should be lowercase with underscores separating words. I'll fix this.

sat/cli/bootprep/input/image.py
haasken-hpe: I guess it just looks off to me without that `pass`. I can remove it though

haasken-hpe:
I've addressed the code review comments, but before you review again, I'd like to make the schema changes discussed during scrum today. That is, instead of this:

```
images:
- name: cos-image
  product:
    name: cos
    version: latest
    is_recipe: true
```

I'd like to create a new `base` property under which either the `ims` key or `product` key will be specified.  I will also replace `is_recipe` with `type: [recipe|image]`, which seems a little more intuitive. Something like this:

```
images:
- name: cos-image
  base:
    product:
      name: cos
      version: "{{ cos.version }}"
      type: recipe
```

I'm going to try to preserve backwards compatibility with this change as well. It should not be too hard to do so.

haasken-hpe:
I've implemented handling of the schema changes in the `BaseInputImage` children classes. I just need to test this on a system now to make sure it still works. I'll report back with results from testing.

haasken-hpe:
I re-tested on vale, and it looks good.

Testing results can be viewed in [a Jira comment here](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1409?focusedCommentId=4940363&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4940363).

sat/cli/bootprep/input/image.py
```diff
@@ -479,3 +497,146 @@ def completed_successfully(self):
         return self.image_create_success and self.image_configure_success
 
     begin = begin_image_create
+
+
+class IMSInputImage(BaseInputImage, ABC):
```
jack-stanek-hpe, line 502: nit: FWIW, I don't think putting `ABC` as a superclass is necessary here, since it's a superclass of `BaseInputImage` too.

`abc.ABC` itself is just a trivial class with the metaclass of `ABCMeta`. See the source: https://github.com/python/cpython/blob/main/Lib/abc.py#L184

In addition, the metaclass is determined by the figuring out a subtype of the metaclasses of the bases and the explicitly specified metaclass. There is some (possibly mind-melting) information here, which is kind of interesting in its own right: https://docs.python.org/3/reference/datamodel.html#determining-the-appropriate-metaclass

sat/cli/bootprep/input/image.py
```diff
@@ -479,3 +497,146 @@ def completed_successfully(self):
...
+    """An input image that specifies a base IMS image/recipe under the 'ims' property under the 'base' property."""
+    @property
+    def ims_data(self):
+        """dict: the data that defines the base IMS image or recipe"""
+        return self.image_data['base']['ims']
+
+    @property
+    def base_is_recipe(self):
+        """bool: whether the starting point is a recipe in IMS or an image"""
+        # is_recipe is a required property
```
jack-stanek-hpe, line 564: nit: I think this is copy/paste cruft, should it instead be the following
```suggestion
        # type is a required property
```

sat/cli/bootprep/input/image.py
```diff
@@ -479,3 +497,146 @@ def completed_successfully(self):
...
+        return self.ims_data['type'] == 'recipe'
+
+
+class ProductInputImage(BaseInputImage):
+    """An input image that specifies a product-provided IMS image or recipe as a starting point."""
+
+    @property
+    def base_is_recipe(self):
+        """bool: whether the starting point is a recipe in IMS or an image"""
+        # is_recipe is a required property
```
jack-stanek-hpe, line 574: nit: Same as above, I think this copy/paste cruft.
```suggestion
        # base.product.type is a required property
```

sat/data/schema/bootprep_schema.yaml
```diff
@@ -34,6 +51,38 @@ description: >
...
+
+      - If the input version has an older minor version than the current
+        version, it may be incompatible, but will be accepted with a warning.
+
+      - If the input version has an older patch version than the current
+        version, it is compatible.
+
+      - If the input version is newer than the current version, it is
+        incompatible because it could contain new fields not understood by the
+        current version.
```
jack-stanek-hpe, line 80: The more I think about this, I'm wondering if we want a similar nuance for when the input version is newer than the current version -- i.e. if just the input patch version is newer, then theoretically the older version of the schema should be compatible, though the validation would probably throw up errors on unrecognized fields. I wonder if there is a option that can be passed to the validator to ignore errors or throw away information from unrecognized fields. This may be outside the scope of this change, and I don't think it's a deal-breaker as is, but I just got to thinking about it more as I looked again.

jack-stanek-hpe (approved):
Looks good, I mostly just had nitpicks.

sat/cli/bootprep/input/image.py
haasken-hpe: Before I added the `ims_data` abstract property to this class, I was getting a warning from PyCharm that said "Class IMSInputImage must implement all abstract methods". This warning is no longer present if I remove the `ABC` parent class now that I have a method on the class with the `abstractmethod` decorator. If I remove that abstract method, the warning comes back. I wonder if this is just a PyCharm limitation.

sat/cli/bootprep/input/image.py
haasken-hpe: Yep, good catch!

sat/data/schema/bootprep_schema.yaml
haasken-hpe: Yeah, good question. Because we have `additionalProperties: false` in most places throughout the schema, I think generally speaking, when we add new properties to the schema, newer input files that use that version are not going to be validate against older schema versions because like you said, there will be errors on unrecognized properties. As far as a way to ignore extra properties, one way would be to stop setting `additionalProperties: false. However, [this article](https://snowplowanalytics.com/blog/2014/05/13/introducing-schemaver-for-semantic-versioning-of-schemas/) I read suggested that disallowing additional properties is a helpful way to ensure that files which validated with older versions of the schema will still validate with the newer version of the schema. See section 2.1 and 2.2 there.

It feels most important to me to support older input files with newer versions of the schema and bootprep code than it is to support newer versions of the input file schema with older versions of the schema and bootprep code. I think this approach handles that more important use case pretty well.

I'll keep this in mind though, and we can discuss further if you'd like.

CHANGELOG.md
```diff
@@ -35,6 +35,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 	- `name` of elements of the `configurations` array
 	- `name`, `branch`, and `version` of elements under `layers` in
 	  elements of the `configurations` array
+- Defined a `sat bootprep` input file schema version and began validating the
+  schema version specified by `sat bootprep` input files.
+- Added functionality to `sat bootprep` to look up images and recipes provided
```
haasken-hpe, line 40: I should probably add something here about the fact that the schema for images changed a bit as well. I'll do that.

sat/cli/bootprep/documentation.py
```diff
@@ -117,7 +117,7 @@ def generate_docs_tarball(schema_path, output_dir):
         with tempfile.TemporaryDirectory() as tempdir:
             with cwd(tempdir), silence():
                 docs_output_path = os.path.join(tempdir, DEFAULT_HTML_INDEX)
-                generate_from_filename(schema_path, docs_output_path)
+                generate_from_filename(schema_path, docs_output_path, deprecated_from_description=True)
```
haasken-hpe, line 120: Oh, I am not using this now. I'll get rid of this change, which was causing strange issues with logging for some reason too.

PR #35, opened 2022-07-29
CRAYSAT-1411: Refactor image dependencies and add name templating

haasken-hpe:
## Summary and Scope
Add the ability to use the name of the base image or recipe that was
used to construct an image as a variable when templating the `name`
property of an image.

Add the ability to use the name of the image as a variable when
templating the `name` property of a session template.

Since names of session templates and images are now templated and not
required to be unique and cannot be rendered until the base is
validated, change the `__str__` methods of both `InputSessionTemplate`
and `BaseInputImage` to simply state the index of the item in the input
instance.

Also add general variable substitution (e.g. product versions) to the
`conifguration` property for images and to the `name`, `image`, and
`configuration` properties of session templates.

Since image names are now templated using variables from the base of the
image, it is necessary to change how images in a `sat bootprep` input
file refer to other images as a base. Add a `ref_name` property on the
images which gives them a unique static name that can be referenced by
another image's `base.image_ref` property. It is no longer supported to
specify a dependency between images using the IMS image name.

Now that we are no longer checking for dependencies between images based
on the images' IMS names, add a check to ensure there is no overlap
between the IMS names of images created by the input file and the IMS
names of images used as bases. This would result in possible race
conditions and non-deterministic behavior if the base of on image is
created by another image in the input file.

## Issues and Related PRs

* Resolves [CRAYSAT-1411](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1411)

### Tested on:

  * vale
  * virtual env on local macbook (for generating and viewing schema docs)

### Test description:

Tested on vale with a bootprep input file that builds an image from the
recipe provided by the CSM product, and then customizes it with a second
image specification that refers to the first by its `ref_name`.

Also tested using the `image.name` variable substitution in BOS session
template names.

## Risks and Mitigations

This does change the schema, but it does so in a backwards-compatible way.

It does also change the behavior of determining dependencies between images. This was not clearly documented before this change anyway, and I don't think users necessarily understood how that would work.

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 	- `name` of elements of the `configurations` array
 	- `name`, `branch`, and `version` of elements under `layers` in
 	  elements of the `configurations` array
+	- `name`, `base.product.version`, and `configuration` properties of
+	  elements in the `images` array.
+	- `name`, `configuration`, and `image` properties of elements in the
+	  `session_templates` array.
 - Defined a `sat bootprep` input file schema version and began validating the
   schema version specified by `sat bootprep` input files.
 - Added functionality to `sat bootprep` to look up images and recipes provided
@@ -43,6 +47,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Changed the `sat bootprep` input file schema by adding a new `base` property
   for an image and moved the existing `ims` property beneath that new property.
+- Changed how `sat bootprep` determines dependencies between images in the input
+  file using new `ref_name` and `base.image_ref` properties.
 
 ### Deprecated
 - Specifying the `ims` property at the top level of an image in the `sat
--- a/sat/cli/bootprep/image.py
+++ b/sat/cli/bootprep/image.py
@@ -30,7 +30,7 @@ import math
 
 from sat.apiclient import APIError, CFSClient, IMSClient
 from sat.cli.bootprep.errors import ImageCreateError
-from sat.cli.bootprep.input.image import ProductInputImage
+from sat.cli.bootprep.input.image import DependentInputImage, IMSInputImage
 from sat.cli.bootprep.public_key import get_ims_public_key_id
 from sat.session import SATSession
 from sat.util import pester_choices
@@ -44,6 +44,24 @@ from sat.waiting import (
 LOGGER = logging.getLogger(__name__)
 
 
+def validate_unique_image_ref_names(input_images):
+    """Validate that all images have unique value for ref_name property.
+
+    Args:
+        input_images (list of sat.cli.bootprep.input.image.BaseInputImage): the
+            input images in the bootprep input instance
+
+    Raises:
+        ImageCreateError: if the images do not have unique ref_name values
+    """
+    counts_by_ref_name = Counter(image.ref_name for image in input_images if image.ref_name)
+    non_unique_ref_names = [name for name, count in counts_by_ref_name.items() if count > 1]
+    if non_unique_ref_names:
+        raise ImageCreateError(f'Images must have unique values for "ref_name" if specified.'
+                               f'Non-unique names: '
+                               f'{", ".join(non_unique_ref_names)}')
+
+
 def validate_unique_image_names(input_images):
     """Validate that the input images all have unique names
 
@@ -51,17 +69,56 @@ def validate_unique_image_names(input_images):
     session templates in a bootprep input file.
 
     Args:
-        input_images (list of sat.cli.bootprep.input.image.InputImage): the input
-            images in the bootprep input instance
+        input_images (list of sat.cli.bootprep.input.image.BaseInputImage): the
+            input images in the bootprep input instance
 
     Raises:
         ImageCreateError: if the images do not have unique names
     """
-    counts_by_name = Counter(image.name for image in input_images)
+    counts_by_name = Counter()
+
+    failed_name_renders = 0
+    for image in input_images:
+        try:
+            counts_by_name[image.name] += 1
+        except ImageCreateError as err:
+            failed_name_renders += 1
+            LOGGER.error(f'Failed to render name of {image}: {err}')
+
+    if failed_name_renders:
+        raise ImageCreateError(f'Failed to render {failed_name_renders} image names.')
+
     non_unique_names = [name for name, count in counts_by_name.items() if count > 1]
     if non_unique_names:
-        raise ImageCreateError('Names of images to create must be unique. Non-unique names: '
+        raise ImageCreateError(f'Names of images to create must be unique. Non-unique names: '
                                f'{", ".join(non_unique_names)}')
+
+
+def validate_no_overwritten_ims_bases(input_images):
+    """Validate that no images will overwrite another's IMS base.
+
+    The supported way for an image to depend on another is to refer to an image
+    by its ref_name in the base.image_ref property. Replacing an IMS image used
+    by another image as its base can result in undefined behavior because both
+    images could be built at the same time.
+    """
+    # Find all named IMS images used as bases
+    name_based_images = [image for image in input_images
+                         if isinstance(image, IMSInputImage) and not image.base_is_recipe
+                         and 'name' in image.ims_data]
+
+    conflicts_exist = False
+    for name_based_image in name_based_images:
+        for other_image in input_images:
+            if other_image.name == name_based_image.ims_data['name']:
+                conflicts_exist = True
+                LOGGER.error(f'{name_based_image} uses {name_based_image.base_description}'
+                             f'which conflicts with the image created by {other_image}.')
+
+    if conflicts_exist:
+        raise ImageCreateError(f'Conflicts between images bases and generated image names '
+                               f'exist. Use "ref_name" and "base.image_ref" to correctly '
+                               f'declare dependencies between images.')
 
 
 def get_ims_images_by_name(ims_client):
@@ -151,6 +208,8 @@ def handle_existing_images(ims_client, input_images, overwrite, skip, dry_run):
         return input_images
     elif skip:
         LOGGER.info(msg_template, {'action': 'skipped'})
+        LOGGER.info('Any images which depend on these images will also be skipped.')
+
         # Create all images that do not already exist
         return [image for image in input_images if image.name not in existing_input_names]
 
@@ -161,28 +220,29 @@ def find_image_dependencies(input_images):
     For example, suppose the bootprep input file looks like this:
 
     images:
-    - name: compute-gpu
-      ims:
-        is_recipe: false
-        name: compute-base
+    - name: "{{ base.name }}"
+      ref_name: cos-base-image
+      base:
+        product:
+          name: cos
+          version: "{{ cos.version }}"
+    - name: compute-gpu-{{ base.name }}
+      base:
+        image_ref: cos-base-image
       configuration: compute-gpu
       configuration_group_names:
       - Compute
-    - name: compute-no-gpu
-      ims:
-        is_recipe: false
-        name: compute-base
+    - name: compute-no-gpu-{{ base.name }}
+      base:
+        image_ref: cos-base-image
       configuration: compute-no-gpu
       configuration_group_names:
       - Compute
-    - name: compute-base
-      ims:
-        is_recipe: true
-        name: compute-recipe
 
-    In this case, the compute-gpu and compute-no-gpu images both depend on the
-    compute-base image being built first from recipe, and then each customizes
-    it with a different configuration.
+    In this case, the second and third images whose names start with "compute-gpu"
+    and "compute-no-gpu", respectively, both depend on the first image with the
+    "ref_name" property value of "cos-base-image". That image must be built first,
+    before the second and third images customize it with different configurations.
 
     Args:
         input_images (list of sat.cli.bootprep.input.image.InputImage): the list
@@ -194,39 +254,41 @@ def find_image_dependencies(input_images):
 
     Raises:
         ImageCreateError: if the images defined in the input file have circular
-            dependencies.
+            dependencies or unresolved dependencies.
     """
-    input_images_by_name = {input_image.name: input_image
-                            for input_image in input_images}
-    cycles_exist = False
-    for image in input_images:
-        # If it is built from a recipe, or specifies an image by its id, or it
-        # is based on an image or recipe provided by a product, then it should
-        # not depend on any named images in the input file.
-        if isinstance(image, ProductInputImage) or image.base_is_recipe or 'id' in image.ims_data:
+    # Get a list of any input images which have dependencies
+    input_images_by_ref_name = {input_image.ref_name: input_image
+                                for input_image in input_images if input_image.ref_name is not None}
+    errors_exist = False
+    for index, image in enumerate(input_images):
+
+        # Only DependentInputImages will depend on other images
+        if not isinstance(image, DependentInputImage):
             continue
 
-        # This is an IMSInputImage, and since 'id' is absent, 'name' must be present
-        base_image_name = image.ims_data['name']
-        if base_image_name in input_images_by_name:
-            LOGGER.info(f'Image named {image.name} depends on image named {base_image_name} '
-                        f'which will also be created by this command.')
-            try:
-                image.add_dependency(input_images_by_name[base_image_name])
-            except DependencyCycleError as err:
-                LOGGER.error(str(err))
-                cycles_exist = True
+        if image.base_image_ref in input_images_by_ref_name:
+            dependency_image = input_images_by_ref_name[image.base_image_ref]
+            LOGGER.info(f'{image} depends on {dependency_image}.')
 
-    if cycles_exist:
-        raise ImageCreateError('Circular dependencies exist in images to be created. '
-                               'Resolve the circular dependencies and try again.')
+            try:
+                image.add_dependency(dependency_image)
+            except DependencyCycleError as err:
+                errors_exist = True
+                LOGGER.error(str(err))
+        else:
+            errors_exist = True
+            LOGGER.error(f'Unable to resolve {image.base_description}.')
+
+    if errors_exist:
+        raise ImageCreateError('Invalid dependencies exist in images to be created. '
+                               'Resolve the dependency issues and try again.')
 
 
 def validate_image_ims_bases(input_images):
     """Validate that the ims bases are valid for the given input images
 
     Args:
-        input_images (list of sat.cli.bootprep.input.image.InputImage): the
+        input_images (list of sat.cli.bootprep.input.image.BaseInputImage): the
             IMSImages which should have their IMS base validated. These images
             should be just the images without dependencies on other images from
             the input, since an image with dependencies depends on images that
@@ -247,7 +309,7 @@ def validate_image_ims_bases(input_images):
             LOGGER.debug(f'IMS base for {image}: {ims_base}')
         except ImageCreateError as err:
             failed_images.append(image)
-            LOGGER.error(f'Failed to find match in IMS for {image.base_description}: {err}')
+            LOGGER.error(f'Failed to find base for {image}: {err}')
 
     if failed_images:
         raise ImageCreateError(f'Failed to find match in IMS for {len(failed_images)} '
@@ -370,23 +432,25 @@ def create_images(instance, args):
     for image in input_images:
         image.public_key_id = ims_public_key_id
 
+    validate_unique_image_ref_names(input_images)
+    # Validate dependencies on the full set of input images
+    find_image_dependencies(input_images)
+
+    # Raises ImageCreateError if validation of IMS bases fails
+    # Do this early to provide information required to render names.
+    validate_image_ims_bases(input_images)
+
+    # Raises ImageCreateError if validation of CFS configurations fails
+    validate_image_configurations(input_images, cfs_client,
+                                  instance.input_configuration_names, args.dry_run)
+
+    # This is where name properties are first accessed and thus rendered
     validate_unique_image_names(input_images)
+
     images_to_create = handle_existing_images(ims_client, input_images, args.overwrite_images,
                                               args.skip_existing_images, args.dry_run)
 
-    # Raises ImageCreateError if validation of CFS configurations fails
-    validate_image_configurations(images_to_create, cfs_client,
-                                  instance.input_configuration_names, args.dry_run)
-
-    create_verb = ('Creating', 'Would create')[args.dry_run]
-    LOGGER.info(f'{create_verb} {len(images_to_create)} images.')
-
-    if not input_images:
-        LOGGER.info('Given input did not define any IMS images')
-        return
-
-    # This can raise ImageCreateError if there are any cycles
-    find_image_dependencies(images_to_create)
+    validate_no_overwritten_ims_bases(images_to_create)
 
     images_without_dependencies = [image for image in images_to_create
                                    if not image.has_dependencies()]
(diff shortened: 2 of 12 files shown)


sat/cli/bootprep/image.py
```diff
@@ -370,22 +432,25 @@ def create_images(instance, args):
...
-        LOGGER.info('Given input did not define any IMS images')
-        return
+    images_to_create = handle_existing_images(ims_client, input_images, args.overwrite_images,
+                                              args.skip_existing_images, args.dry_run)
+
+    validate_no_overwritten_ims_bases(images_to_create)
 
-    # This can raise ImageCreateError if there are any cycles
+    # This can raise ImageCreateError for unresolvable dependencies
     find_image_dependencies(images_to_create)
```
haasken-hpe, line 454: I think this should actually be done before `handle_existing_images`. Suppose you have image A and B in the input file, and B depends on A. Let's say A already exists, and the admin chose to skip creating it. Then `images_to_create` would not contain image A. Then when `images_to_create` is passed to `find_image_dependencies`, B's dependency on A would be unresolvable.

I think it's safe to move this up since it doesn't depend on any names being rendered or anything. It would probably make sense to do right after `validate_unique_image_ref_names`. I'm testing this now.

haasken-hpe: Here's an example of the bug:

```
ncn-m001:~/haasken/sprint-reviews/latest # sat bootprep run --recipe-version latest barebones-customized.yaml
INFO: Validating given input file barebones-customized.yaml
INFO: Input file successfully validated against schema
INFO: Creating 1 CFS configuration(s)
The following CFS configurations already exist: haasken-test-22.07.1. Would you like to skip, overwrite, or abort? [skip,overwrite,abort] skip
INFO: The following CFS configurations already exist and will be skipped: haasken-test-22.07.1
WARNING: Found 2 public keys in IMS whose contents match the contents of public key file /root/.ssh/id_rsa.pub. Using the first one, but you may wish to clean up the duplicates.
INFO: Using IMS public key with id 12873d8a-051f-4ef0-9f1e-324a6eb5112c
INFO: Found IMS base for image at index 0: recipe provided by latest version of product csm
INFO: Found IMS base for image at index 1: image from input instance with ref_name="haasken-barebones"
One or more images already exist in IMS with the following names: haasken-cray-shasta-csm-sles15sp3-barebones.x86_64-csm-1.2. Would you like to skip, overwrite, or abort? [skip,overwrite,abort] skip
INFO: Images with the following names already exist in IMS and will be skipped: haasken-cray-shasta-csm-sles15sp3-barebones.x86_64-csm-1.2
INFO: Any images which depend on these images will also be skipped.
ERROR: Unable to resolve image from input instance with ref_name="haasken-barebones".
ERROR: Invalid dependencies exist in images to be created. Resolve the dependency issues and try again.
```

Here is after the re-ordering:

```
(c85af60ff17d) sat-container:/sat/share # sat bootprep run --recipe-version latest barebones-customized.ya
ml
INFO: Validating given input file barebones-customized.yaml
INFO: Input file successfully validated against schema
INFO: Creating 1 CFS configuration(s)
The following CFS configurations already exist: haasken-test-22.07.1. Would you like to skip, overwrite, or abort? [skip,overwrite,abort] skip
INFO: The following CFS configurations already exist and will be skipped: haasken-test-22.07.1
WARNING: Found 2 public keys in IMS whose contents match the contents of public key file /root/.ssh/id_rsa.pub. Using the first one, but you may wish to clean up the duplicates.
INFO: Using IMS public key with id 12873d8a-051f-4ef0-9f1e-324a6eb5112c
INFO: image at index 1 depends on image at index 0.
INFO: Found IMS base for image at index 0: recipe provided by latest version of product csm
INFO: Found IMS base for image at index 1: image from input instance with ref_name="haasken-barebones"
One or more images already exist in IMS with the following names: haasken-cray-shasta-csm-sles15sp3-barebones.x86_64-csm-1.2. Would you like to skip, overwrite, or abort? [skip,overwrite,abort] skip
INFO: Images with the following names already exist in IMS and will be skipped: haasken-cray-shasta-csm-sles15sp3-barebones.x86_64-csm-1.2
INFO: Any images which depend on these images will also be skipped.
INFO: Of the 1 that will be created, 0 have no dependencies and will be created first.
INFO: Creating 1 images.
INFO: Creating images
INFO: Image creation completed successfully
INFO: Creating 1 BOS session template
```

CHANGELOG.md
```diff
@@ -35,6 +35,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 	- `name` of elements of the `configurations` array
 	- `name`, `branch`, and `version` of elements under `layers` in
 	  elements of the `configurations` array
+	- `name`, `base.product.version`, and `configuration` properties of
```
jack-stanek-hpe, line 38: I think it might be good to specify what the "special" variable names are (e.g. `recipe`, `image`, `base`, etc.) for the user's benefit. We don't have to do that here in the changelog, but it's probably valuable to document that somewhere. I also think we probably want to enforce somewhere that those variables can't be defined in the variables file or on the command line. (e.g. can't do something like `--vars base="something"`)

sat/cli/bootprep/image.py
```diff
@@ -44,26 +44,83 @@
...
+            failed_name_renders += 1
+            LOGGER.error(f'Failed to render name of {image}: {err}')
+
+    if failed_name_renders:
+        raise ImageCreateError(f'Failed to render {failed_name_renders} image names.')
+
     non_unique_names = [name for name, count in counts_by_name.items() if count > 1]
     if non_unique_names:
-        raise ImageCreateError('Names of images to create must be unique. Non-unique names: '
+        raise ImageCreateError(f'Names of images to create must be unique. Non-unique names: '
```
jack-stanek-hpe, line 93: nit: FWIW, I don't think the `f` is actually necessary since there aren't any variable substitutions on this line. See [PEP-498](https://peps.python.org/pep-0498/#concatenating-strings).

sat/cli/bootprep/input/base.py
```diff
@@ -38,23 +38,109 @@
...
+            result = func(self, *args, **kwargs)
+            self.jinja_context[context_var or func.__name__] = result
+            return result
+
+        return wrapper
+
+    return decorator
+
+
+def jinja_rendered(func):
```
jack-stanek-hpe, line 72: Substituting the `recipe` variable in templates fails with a "variable undefined" error when the `--recipe-version` argument isn't passed, which in my opinion doesn't give much intuition as to what's causing the issue -- when I encountered it, I assumed it was a bug in SAT, not an error in usage. To address this, I created a [commit to address this in my personal fork](https://github.com/jack-stanek-hpe/sat/commit/a41b53f538db8bc7cd813763d67d8b576544a149). It prints a special error message when the `recipe` variable is undefined and `--recipe-version` isn't present, but returns the default error in all other cases.

jack-stanek-hpe (commented):
I think the errors brought up in my review, particularly regarding issues referencing images in the input file, are sufficient for us to postpone merging this PR until Ryan returns. Outside of the comments, I think the PR looks mostly good.

sat/cli/bootprep/input/session_template.py
```diff
@@ -95,6 +112,7 @@ def image_is_uuid(self):
         return bool(uuid_regex.fullmatch(self.image))
 
     @cached_property
+    @provides_context('image')
```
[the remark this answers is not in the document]
haasken-hpe: Still working on understanding this bug and considering your fix and how it interacts with Eli's fix proposed in #37.

As discussed in scrum today, I noticed that your example is creating an image from a base image, but not configuring it, and I noticed that `sat bootprep` does not handle that very well. I filed a [CRAYSAT-1529](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1529) for that separate problem.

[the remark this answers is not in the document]
haasken-hpe: My original goal with this change was to encourage users to switch to using the `ref_name` to refer to images from the input file in their session templates (as implemented in CRAYSAT-1410 PR #37). If the user is templating their image names in the input file, referring to those images by a non-templated `ref_name` is really the only way to achieve that anyway.

However, you have a valid point here. People may have older input files that use this old way of referencing images from the input file (by their non-templated names) in BOS session templates. If it's easy to continue to support that, we probably should.

I'll just add that this problem appears to be specific to performing a dry-run (or validation) against the input file. If you actually run `sat bootprep` in a non-dry-run mode against such an input file, it will build the IMS image (ignoring CRAYSAT-1529 for now), and then when the session templates are validated, it will find the named image in IMS, and will successfully create the BOS session template. This is not an argument against fixing the dry-run/validation case, but I just wanted to make the distinction clear. I do actually want to expose `validate` as a separate action still, e.g. `sat bootprep validate input.yaml` (see [CRAYSAT-1284](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1284))

With respect to your fix, I am okay with the overall idea. I actually briefly had an implementation similar to this where I would set the `image` variable in the context to a dictionary with just the 'name' key if the image was from the input instance. I think I'm going to do that instead of setting the `image` variable in the context to a `BaseInputImage` object. I think I will also limit the information provided in the `image` context variable to just the 'name' when the IMS record is pulled. We can then consider if more of the info provided by IMS is needed. For now, I think it's better to keep the info provided in that context the same between the two cases.

Such a change will require a little bit of change to Eli's commit added to CRAYSAT-1410, but it should not be too involved. We will basically just need to make sure we handle setting the `image` variable context appropriately in both `InputSessionTemplate` classes' `validate_image_exists` logic.

sat/cli/bootprep/input/image.py
```diff
@@ -250,6 +270,10 @@ def begin_image_create(self):
 
         LOGGER.info(f'Launching IMS job to create image')
 
+        ims_base = self.ims_base
```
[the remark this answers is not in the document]
haasken-hpe: I do believe this is an uncommon use case, and I believe it's actually a pre-existing bug. It would be handled by [CRAYSAT-1529](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1529). I am going to add your comment here to that Jira.

sat/cli/bootprep/input/base.py
haasken-hpe: Neat to see an example of using the `undefined` kwarg 👍 

However, I should point out that `--recipe-version` is not actually what defines `recipe.version`. The `--recipe-version` flag will just tell `sat bootprep` which version of the recipe to go look up in VCS to find variables. Then there is a `recipe` key in the `product_vars.yaml` file. Maybe the recipe version should just come from the version of the recipe we specify or found in VCS based on existing branches. Also, I think all the other variables would give similarly confusing errors if you didn't have a `product_vars.yaml` file to look up.

I also originally wanted to make `--recipe-version` default to the latest version. I think that would resolve this issue in another way, would you agree?

sat/cli/bootprep/image.py
haasken-hpe: I tend to just do `f` on all parts of concatenated strings when I have any need for it in the overall concatenated strings for two reasons:

* In case somebody adjusts the line length
* In case somebody adds a variable and doesn't realize each part isn't an f-string.
* For visual consistency with open single-quote chars being aligned

CHANGELOG.md
haasken-hpe: Agree. This is a gap. I filed [CRAYSAT-1530](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1530) for this.

FWIW, I think if you did try `--vars base="something"`, it would just be ignored in cases where `base` has a special meaning, i.e. in image names. I'll try to be clear about that in docs.

sat/cli/bootprep/input/base.py
jack-stanek-hpe: Yes, that would solve it in a different way, though the problem still exists in the case of the hpc-software-recipe repo not existing or if there's some error when fetching it -- in that case, we may not want to blanket fail in every case, since there's not theoretical reason that an input file which doesn't reference the recipe vars should fail if the recipe can't be fetched. In such a case, a "progressive enhancement" style paradigm could potentially suit both cases and cover the errors more gracefully, e.g. where the recipe vars are fetched on-demand when the recipe calls for them.

jack-stanek-hpe: Another thought I had w.r.t. variables being pulled in from the software recipe, which relates to [CRAYSAT-1530](CRAYSAT-1530), is that maybe it could be useful to dynamically document which variables are available given some context, in addition to what referential variables exist. For instance, given some recipe, the variables defined within that recipe could be listed out, perhaps using another subcommand.

haasken-hpe: > Another thought I had w.r.t. variables being pulled in from the software recipe, which relates to [CRAYSAT-1530](CRAYSAT-1530), is that maybe it could be useful to dynamically document which variables are available given some context, in addition to what referential variables exist. For instance, given some recipe, the variables defined within that recipe could be listed out, perhaps using another subcommand.

Interesting idea. I opened [CRAYSAT-1532](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1532) for an option or action to list recipe variables.

haasken-hpe: > Yes, that would solve it in a different way, though the problem still exists in the case of the hpc-software-recipe repo not existing or if there's some error when fetching it -- in that case, we may not want to blanket fail in every case, since there's not theoretical reason that an input file which doesn't reference the recipe vars should fail if the recipe can't be fetched. In such a case, a "progressive enhancement" style paradigm could potentially suit both cases and cover the errors more gracefully, e.g. where the recipe vars are fetched on-demand when the recipe calls for them.

I agree that it would be nice not to fail if the input file wasn't using the variables from the HPC software recipe. There's actually a comment in this method about changing an error to a warning when we default to the latest version of a recipe: https://github.com/Cray-HPE/sat/blob/9c509fa852e901efe60e2d11be4f5a5ec913130a/sat/cli/bootprep/vars.py#L115-L143

I think I had it implemented as a warning initially when I had it always defaulting to the latest software recipe version. Once I changed it so that the user had to specify a `--recipe-version`, I made it an error because in that case, the user would have had to explicitly ask for recipe variables.

haasken-hpe: I'm going to change it to always attempt to load the latest HPC software recipe variables from VCS, and if it fails, I'm going to log a warning that rendering of templates which use those variables may fail.  That is, `--recipe-version` will not be required. I think that will improve usability significantly for the typical admin. Let me know if you disagree.

Being able to dynamically determine whether an input file is trying to use any recipe variables seems nice, but I'm not sure how we'd do that without knowing which variables we would expect a software recipe to define. I think it's probably good enough to at least default to the latest recipe version and only warn if there is a failure to load it.

ekamin-hpe (approved):

sat/cli/bootprep/input/base.py
```diff
@@ -252,19 +355,36 @@ def validate_items(self, **kwargs):
...
+    # validation will render names using Jinja2 templates.
+    @Validatable.validation_method(ordinal=1)
     def validate_unique_names(self, **_):
         """Validate that all items in collection have unique names.
 
         Raises:
             InputItemValidateError: if there are any items with the same name
         """
-        counts_by_name = Counter(item.name for item in self.items)
+        counts_by_name = Counter()
```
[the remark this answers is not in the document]
haasken-hpe: Other code does depend on rendering the names here as well, so I do still want to do that at least.

sat/cli/bootprep/input/session_template.py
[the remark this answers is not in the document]
haasken-hpe: Resolved by e8ee3aad37b9442b4e74e1094a378deeaaeabf99

sat/cli/bootprep/input/base.py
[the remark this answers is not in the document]
ekamin-hpe: Sounds good, I had not thought about the rendering order.

sat/cli/bootprep/input/session_template.py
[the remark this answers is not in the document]
haasken-hpe: Here's a test showing that this problem is resolved (as long as the image isn't a reference to an unchanged existing IMS image; see CRAYSAT-1529).

First, the input file:

```
ncn-m001:~/haasken/bootprep # cat jstanek-comment.yaml
---
images:
- name: haasken-barebones-image
  base:
    ims:
      name: cray-shasta-csm-sles15sp3-barebones.x86_64-csm-1.2
      type: recipe

session_templates:
- name: compute-{{image.name}}
  image: haasken-barebones-image
  configuration: cos-config-27.06
  bos_parameters:
    boot_sets:
      compute:
        node_roles_groups:
        - "Compute"
```

Then running in dry-run mode:

```
ncn-m001:~/haasken/bootprep # sat bootprep run --dry-run jstanek-comment.yaml
INFO: Validating given input file jstanek-comment.yaml
INFO: Input file successfully validated against schema
INFO: Given input did not define any CFS configurations
WARNING: Found 2 public keys in IMS whose contents match the contents of public key file /root/.ssh/id_rsa.pub. Using the first one, but you may wish to clean up the duplicates.
INFO: Using IMS public key with id 12873d8a-051f-4ef0-9f1e-324a6eb5112c
INFO: Found IMS base for image at index 0: recipe with name=cray-shasta-csm-sles15sp3-barebones.x86_64-csm-1.2
One or more images already exist in IMS with the following names: haasken-barebones-image. Would you like to skip, overwrite, or abort? [skip,overwrite,abort] overwrite
INFO: Images with the following names already exist in IMS and would be overwritten: haasken-barebones-image
INFO: Of the 1 that would be created, 1 have no dependencies and would be created first.
INFO: Would create 1 images.
INFO: Dry run, not creating images.
```

I also ran in non-dry-run mode, and skipped the existing image:

```
ncn-m001:~/haasken/bootprep # sat bootprep run --skip-existing-images jstanek-comment.yaml
INFO: Validating given input file jstanek-comment.yaml
INFO: Input file successfully validated against schema
INFO: Given input did not define any CFS configurations
WARNING: Found 2 public keys in IMS whose contents match the contents of public key file /root/.ssh/id_rsa.pub. Using the first one, but you may wish to clean up the duplicates.
INFO: Using IMS public key with id 12873d8a-051f-4ef0-9f1e-324a6eb5112c
INFO: Found IMS base for image at index 0: recipe with name=cray-shasta-csm-sles15sp3-barebones.x86_64-csm-1.2
INFO: Images with the following names already exist in IMS and will be skipped: haasken-barebones-image
INFO: Any images which depend on these images will also be skipped.
INFO: Of the 0 that will be created, 0 have no dependencies and will be created first.
INFO: Creating 0 images.
INFO: Creating images
INFO: Image creation completed successfully
INFO: Creating 1 BOS session template
ncn-m001:~/haasken/bootprep # cray bos v1 sessiontemplate list | name_contains haasken
[
  {
    "boot_sets": {
      "compute": {
        "etag": "a2c298be8246cdf182a277ea1c61f13c",
        "node_roles_groups": [
          "Compute"
        ],
        "path": "s3://boot-images/be05add1-c3a5-4043-bcbd-08ae3e90b124/manifest.json",
        "rootfs_provider": "cpss3",
        "rootfs_provider_passthrough": "dvs:api-gw-service-nmn.local:300:nmn0",
        "type": "s3"
      }
    },
    "cfs": {
      "configuration": "cos-config-27.06"
    },
    "enable_cfs": true,
    "name": "compute-haasken-barebones-image"
  }
]
```

sat/cli/bootprep/input/base.py
haasken-hpe: I made this change with 0b8fb2d6346d3dc1a0435a98636a7816116b4d3c.

I also filed [CRAYSAT-1533](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1533) to get the recipe version from the branch name rather than requiring it to be a variable in the file. This will require work here and in hpc-shasta-software-recipe.

haasken-hpe: Oops, there's a bug in that commit which results in this:

```
ncn-m001:~/haasken/sprint-reviews/latest # sat bootprep run --dry-run --save-files --no-resolve-branches --recipe-version 22.07.2 bootprep-cos-config.yaml
INFO: Validating given input file bootprep-cos-config.yaml
INFO: Input file successfully validated against schema
WARNING: Failed to load variables defined by HPC software recipe version 22.07.2: No recipe with version 22.07.2 is installed on the system.
Traceback (most recent call last):
  File "/sat/venv/bin/sat", line 33, in <module>
    sys.exit(load_entry_point('sat==3.17.1', 'console_scripts', 'sat')())
  File "/sat/venv/lib/python3.9/site-packages/sat/main.py", line 120, in main
    subcommand(args)
  File "/sat/venv/lib/python3.9/site-packages/sat/cli/bootprep/main.py", line 266, in do_bootprep
    do_bootprep_run(schema_validator, args)
  File "/sat/venv/lib/python3.9/site-packages/sat/cli/bootprep/main.py", line 185, in do_bootprep_run
    var_context.load_vars()
  File "/sat/venv/lib/python3.9/site-packages/sat/cli/bootprep/vars.py", line 79, in load_vars
    deep_update_dict(self.vars, self.get_hpc_software_recipe_vars())
  File "/sat/venv/lib/python3.9/site-packages/sat/util.py", line 538, in deep_update_dict
    raise TypeError(f'Expected two dict instances. Got {type(original_dict)} and {type(new_dict)}')
TypeError: Expected two dict instances. Got <class 'dict'> and <class 'NoneType'>
```

I just need to return the empty dict in `get_hpc_software_recipe_vars` if there is an error loading the recipe vars.

haasken-hpe: Pushed the fix b8ffde97553d3c3fdf086214dd3c1e83727669bc

New output for this test case:

```
(27dcf023d817) sat-container:/sat/share # sat bootprep run --dry-run --save-files --no-resolve-branches --recipe-version 22.07.2 bootprep-cos-config.yaml
INFO: Validating given input file bootprep-cos-config.yaml
INFO: Input file successfully validated against schema
WARNING: Failed to load variables defined by HPC software recipe version 22.07.2: No recipe with version 22.07.2 is installed on the system.
INFO: Would create 1 CFS configuration(s)
ERROR: Failed to render Jinja2 template haasken-cos-config-{{recipe.version}} for value name: 'recipe' is undefined
```

haasken-hpe:
I just rebased on latest feature/CRAYSAT-1385, which pulled in my commit to fix the name of the hpc-shasta-software-recipe VCS repo and Jack's commit to switch to `packaging.version`. I'll do a quick sanity test of the built docker image before merge.

haasken-hpe:
Tested against a bootprep file with the following contents, and everything was created as expected:

```
---
configurations:
- name: "haasken-test-{{ recipe.version }}"
  layers:
  - name: haasken-test
    git:
      url: https://api-gw-service-nmn.local/vcs/cray/haasken-config-management.git
      branch: master
    playbook: test.yml

images:
- name: "haasken-{{ base.name }}"
  ref_name: haasken-barebones
  base:
    product:
      name: csm
      type: recipe

- name: "configured-{{ base.name }}"
  base:
    image_ref: haasken-barebones
  configuration: "haasken-test-{{ recipe.version }}"
  configuration_group_names:
  - Compute

session_templates:
- name: "{{ image.name[:18] }}-{{ recipe.version }}"
  # Note, this has to be updated depending on the system on which this demo is run
  image: ead73bae-d195-4dda-873a-3611bfe4f747
  configuration: "haasken-test-{{ recipe.version }}"
  bos_parameters:
    boot_sets:
      compute:
        node_roles_groups:
        - Compute
```

PR #41, opened 2022-08-10
CRAYSAT-1488: Replace `distutils` version handling with `packaging`

jack-stanek-hpe:
## Summary and Scope

This change removes the deprecated `distutils` version handling features
and replaces them with the `packaging` library. HPC Software Version
handling and schema version validation were affected.

## Issues and Related PRs

* Resolves [CRAYSAT-1488](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1488)

### Tested on:

  * `vale`

### Test description:

Run unit tests. Test on vale with `sat bootprep run --recipe-version
latest` and ensure that the 22.07 recipe is treated as "latest" over an
example 22.07rc1 recipe.

## Risks and Mitigations

N/A

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated

### DIFF
--- a/requirements.lock.txt
+++ b/requirements.lock.txt
@@ -28,6 +28,7 @@ marshmallow-enum==1.5.1
 mypy-extensions==0.4.3
 natsort==8.1.0
 oauthlib==3.2.0
+packaging==21.3
 paramiko==2.9.2
 parsec==3.5
 prettytable==0.7.2
--- a/sat/cli/bootprep/validate.py
+++ b/sat/cli/bootprep/validate.py
@@ -24,12 +24,12 @@
 """
 Validation functions for the bootprep input file based on the schema.
 """
-from distutils.version import StrictVersion
 import logging
 import pkgutil
 
 from jsonschema import SchemaError
 from jsonschema.validators import validator_for
+from packaging import version
 from yaml import safe_load, YAMLError
 
 from sat.cli.bootprep.constants import DEFAULT_INPUT_SCHEMA_VERSION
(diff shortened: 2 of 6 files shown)


requirements.lock.txt
```diff
@@ -28,6 +28,7 @@ marshmallow-enum==1.5.1
 mypy-extensions==0.4.3
 natsort==8.1.0
 oauthlib==3.2.0
+packaging==21.3
```
ekamin-hpe, line 31: Can you also add this to `requirements-dev.lock.txt` too, just for consistency?

sat/cli/bootprep/validate.py
```diff
@@ -116,12 +116,17 @@ def validate_instance_schema_version(instance, schema_validator):
...
     try:
-        strict_instance_version = StrictVersion(instance_version)
-    except ValueError as err:
+        instance_version = version.parse(instance_version)
+
+        # LegacyVersion is deprecated (see https://github.com/pypa/packaging/issues/321),
+        # and sometimes invalid versions can parse as "Legacy". To avoid this,
+        # treat LegacyVersions as invalid.
+        if isinstance(instance_version, version.LegacyVersion):
+            raise version.InvalidVersion()
```
ekamin-hpe, line 128: nit: unused parens here (I think?)

ekamin-hpe (approved):
LGTM

sat/cli/bootprep/validate.py
```diff
@@ -116,12 +116,17 @@ def validate_instance_schema_version(instance, schema_validator):
...
     # No default is needed here because the version property was added to the
     # bootprep_schema.yaml schema file simultaneously with this code.
-    current_version = schema_validator.schema['version']
-    strict_current_version = StrictVersion(current_version)
+    current_version = version.parse(schema_validator.schema['version'])
 
     try:
-        strict_instance_version = StrictVersion(instance_version)
-    except ValueError as err:
+        instance_version = version.parse(instance_version)
```
haasken-hpe, line 122: This is pretty minor, but the reason I kept `instance_version` and `strict_instance_version` as separate vars was so that the exact string specified under the `schema_version` property in the input instance would be printed to the user in `err_template`. I think once it's parsed, the string representation can look a little different (e.g. zero padding removed).

sat/recipe.py
```diff
@@ -63,13 +63,13 @@ def __init__(self, version, vcs_repo, vcs_branch):
         """Create a new HPCSoftwareRecipe object.
 
         Args:
-            version (str): the recipe version string
+            version (packaging.version.Version): the recipe version string
```
haasken-hpe, line 66: it looks like it should still be a `str` as an arg, but gets set to a `packaging.version.Version` in the attribute.

haasken-hpe (approved):
Very minor comments.

sat/cli/bootprep/validate.py
jack-stanek-hpe: The parens are used since it's calling the constructor for `InvalidVerision()`.

sat/recipe.py
jack-stanek-hpe: Ah, I misread `Args` as `Attributes`. I'll change that back.

sat/cli/bootprep/validate.py
ekamin-hpe: whoops, sounds good.

sat/cli/bootprep/validate.py
jack-stanek-hpe: That's a good point. I can keep the instance version string around and use that in error messages instead.

requirements.lock.txt
jack-stanek-hpe: https://github.com/Cray-HPE/sat/blob/feature/CRAYSAT-1385/requirements-dev.lock.txt#L37

sat/cli/bootprep/validate.py
haasken-hpe: The extra parens don't bother me (in fact I kinda like being explicit), but FWIW:

> The sole argument to [raise](https://docs.python.org/3/reference/simple_stmts.html#raise) indicates the exception to be raised. This must be either an exception instance or an exception class (a class that derives from [Exception](https://docs.python.org/3/library/exceptions.html#Exception)). If an exception class is passed, it will be implicitly instantiated by calling its constructor with no arguments:

https://docs.python.org/3/tutorial/errors.html#raising-exceptions

So if it derives from `Exception`, we don't need the parens.

haasken-hpe (approved):

sat/cli/bootprep/validate.py
jack-stanek-hpe: Huh, that's good to know!

requirements.lock.txt
ekamin-hpe: Oh, I guess so . Looks to me like it is actually a dependency of Sphinx (possibly others as well), which explains why it was in `requirements-dev.lock.txt` but not in `requirements.lock.txt`. Looks like they are the same version, which is reassuring as well.

PR #48, opened 2022-08-25
CRAYSAT-1536: Remove image dependencies when skipped

haasken-hpe:
## Summary and Scope

When an image depends on another image being created by the input file,
and that latter image already exists and re-creation of that image is
skipped, remove that latter image as a dependency of the former image.
This ensures that all images in the input file will be created and exist
by the time we get to the point of creating BOS session templates, which
may reference those images.

Add a `skip` attribute to `BaseInputImage` and set it to true if the
user requests to skip the re-creation of an existing image. Use this
attribute in the computation of the `ims_base` property of the
`DependentInputImage`, so that we can find the ID of the base image when
that dependency image already existed and was not created by this
invocation of `sat bootprep`.

Fix IMS resource cache clearing bug that was exposed by the other changes.
Now when an image is skipped, images are queried from IMS and found by
name earlier in finding the IMS base for the images which depend on the
skipped image.

## Issues and Related PRs

* Resolves [CRAYSAT-1536](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1536)

### Tested on:

  * loki

### Test description:

Created a bootprep input file that defines three images. The first
image is built from the recipe provided by the COS product. The second
image uses the first image as a base and customizes it for computes. The
third image uses the first image as a base as well but customizes it
differently for UAN. The configurations used were not real CN/UAN
configurations, but rather simple CFS configurations with a debug
playbook. The bootprep file also contained two session templates, one
that referenced the CN image and one that referenced the UAN image.

Ran `sat bootprep run` against this input file with the CFS
configuration designed to fail. Ran `sat bootprep run` against this
input file again with the CFS configuration designed to succeed.
Answered the prompt about the existing image with "skip", so the base
image would not be created again. Confirmed that the other two images
were still created, and creation of the BOS session templates succeeded.

## Risks and Mitigations

No known risks.

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated


sat/cli/bootprep/image.py
```diff
@@ -208,7 +208,15 @@ def handle_existing_images(ims_client, input_images, overwrite, skip, dry_run):
...
     elif skip:
         LOGGER.info(msg_template, {'action': 'skipped'})
-        LOGGER.info('Any images which depend on these images will also be skipped.')
+
+        # Remove already existing images as dependencies of other images. They can be built right away
+        for existing_image in existing_input_images:
+            existing_image.skip = True
+            for image in input_images:
+                if existing_image in image.dependencies:
+                    LOGGER.info(f'Removing skipped, already existing {existing_image} '
```
ekamin-hpe, line 217: I feel like this wording is a little bit awkward but I can't think of a better suggestion, wouldn't hold up the merge for it.

ekamin-hpe (approved):

haasken-hpe:
Here's my testing output: https://gist.github.hpe.com/ryan-haasken/f14b6dac9248935b2ba1f2ee7e40cef5

sat/cli/bootprep/image.py
haasken-hpe: I think this message is not super important to the user, so I will drop it to debug level.

PR #49, opened 2022-08-25
CRAYSAT-1546: Fix traceback on invalid subcommand

ekamin-hpe:
This commit removes setting `exit_on_error` to `False` in the
`parse_known_args` method of `sat.parser.SATArgParser`. Add
a change log entry.

Test Description:
* I installed SAT into a virtualenv with Python 3.9
* I ran SAT with an invalid subcommand and verified that a succinct
  error message was printed.
* I ran SAT with a valid subcommand but an invalid option and verified
  that the help text for the subcommand was printed.
* I ran SAT with both an invalid subcommand and invalid option and
  verified it behaved the same as just giving the invalid subcommand.

## Issues and Related PRs

* Resolves [CRAYSAT-1546](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1546)

### Tested on:

### Test description:

See commit message.

## Pull Request Checklist

- [X] Copyrights updated
- [X] License file intact
- [X] Target branch correct
- [X] CHANGELOG.md updated

### DIFF
--- a/sat/parser.py
+++ b/sat/parser.py
@@ -1,7 +1,7 @@
 #
 # MIT License
 #
-# (C) Copyright 2019-2021 Hewlett Packard Enterprise Development LP
+# (C) Copyright 2019-2022 Hewlett Packard Enterprise Development LP
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
 # copy of this software and associated documentation files (the "Software"),
@@ -79,10 +79,6 @@ class SATArgParser(ArgumentParser):
 
         else:
             return parsed
-
-    def parse_known_args(self, args=None, namespace=None):
-        self.exit_on_error = False
-        return super().parse_known_args(args=args, namespace=namespace)
 
     def error(self, message):
         """Prints errors based on invalid arguments.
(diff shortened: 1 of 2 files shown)


ekamin-hpe:
Output from testing here: https://gist.github.hpe.com/eli-kamin/32fbb9d4378de699ac1c6245c3174843

sat/parser.py
```diff
@@ -81,7 +81,6 @@ def parse_args(self, args=None, namespace=None):
             return parsed
 
     def parse_known_args(self, args=None, namespace=None):
-        self.exit_on_error = False
         return super().parse_known_args(args=args, namespace=namespace)
```
haasken-hpe, line 84: It looks like this method is now no different than the version from the parent class, `ArgumentParser`. Would it be equivalent if we just removed this method definition entirely now?

haasken-hpe (approved):

jack-stanek-hpe (approved):

sat/parser.py
ekamin-hpe: Looks like it, I'll go ahead and do that instead, thanks.

haasken-hpe (approved):

PR #38, opened 2022-08-08
PE-42104 - Adding Jobstat to SAT CLI

edzaka-hpe:
## Summary and Scope

New feature named "Jobstat," which allows users and admins to get a system-wide view of application status data via the SAT toolkit.  The change is backwards compatible as long as it has access to the elasticsearch api. The feature is meant to be implemented on any system running PALS. Currently, it is being developed for Slice.

## Issues and Related PRs
PE-42104 - Adding Jobstat to SAT CLI

## Testing
Feature was tested in a pyenv environment. Output of the command gave desired table to the user. 

### Tested on:
  * Local development environment
  * Virtual Shasta

### Test description:

Test included running `sat jobstat` in a virtual environment. The script sent a HTTP GET request to a flask server (which will soon be held in Slice in a Kubernetes pod running on PBSPro). The flask server returns a JSON object from the elasticsearch api via a python script on the pod. `sat jobstat` returns proper table output. Exceptions were tested and handled properly by the script as well.

## Risks and Mitigations
A risk is the elasticsearch api not working with the command.

## Pull Request Checklist

- [x ] Version number(s) incremented, if applicable
- [ x] Copyrights updated
- [ x] License file intact
- [ x] Target branch correct
- [x ] CHANGELOG.md updated
- [x ] Testing is appropriate and complete, if applicable
- [ ] [HPC Product Announcement](https://cray.slack.com/archives/C026TVCSXLH) prepared, if applicable

PR #58, opened 2022-09-13
PE-42104 - Adding Jobstat to SAT CLI (continuation of PR-38)

vinaybabug-hpe:
This commit implements `sat jobstat`. This command gets a system-wide view of
  applications and displays it in a pretty table format. This commit also adds
  a change log entry and man page for the command.

### DIFF
--- a/docs/man/sat-jobstat.8.rst
+++ b/docs/man/sat-jobstat.8.rst
@@ -1,0 +1,60 @@
+=============
+ SAT-JOBSTAT
+=============
+
+---------------------------------------------------
+Jobstat - Check status of jobs and its applications
+---------------------------------------------------
+
+:Author: Hewlett Packard Enterprise Development LP.
+:Copyright: Copyright 2022 Hewlett Packard Enterprise Development LP.
+:Manual section: 8
+
+SYNOPSIS
+========
+
+**sat** [global-opts] **jobstat** [options]
+
+DESCRIPTION
+===========
+
+Jobstat is a command that allows a user or admin to access application and job
+data via the command line. The command provides a table showing the user the
+apid, jobid, user, state, and time-reported of all jobs on the system.
+
+OPTIONS
+=======
+
+These options must be specified after the subcommand.
+
+**-h, --help**
+        Print the help message for 'sat jobstat'.
+
+.. include:: _sat-format-opts.rst
+
+EXAMPLES
+========
+
+The command displays a summary report for all jobs on the system e.g.
+
+::
+
+    # sat jobstat --filter state=completed
+    +---------------------------------------+--------------+-------+-------------------------------------+-----------+-----------+---------------------+
+    | apid                                  | jobid        | user  | command                             | state     | num_nodes | node_list           |
+    +---------------------------------------+--------------+-------+-------------------------------------+-----------+-----------+---------------------+
+    | 7677b7d3-ec35-4e23-8678-92b823347939  | 49.pbs-host  | 5827  | mpiexec --ppn=1 -n2 ./signals       | completed | 2         | nid000001,nid000002 |
+    | 76b783af-5c7e-4462-97c0-3dfd0f04b9b3  | 51.pbs-host  | 5827  | mpiexec --ppn=1 -n2 ./signals       | completed | 2         | nid000001,nid000002 |
+    | 7a85c000-05a9-40cd-8551-00092cead801  | 48.pbs-host  | 5827  | mpiexec -n2 ./signals               | completed | 1         | nid000001           |
+    | 8b5b14b3-8215-4cbe-bba4-7d7c82d48fef  | 50.pbs-host  | 5827  | mpiexec --ppn=1 -n2 ./signals       | completed | 2         | nid000001,nid000002 |
+    | 957bbd4c-de3c-424d-99d6-fd2b728b2941  | 54.pbs-host  | 5827  | mpiexec --ppn=1 -n2 ./signals       | completed | 2         | nid000001,nid000002 |
+    | ba977b71-40b6-40db-aa20-a583edc18be3  | 59.pbs-host  | 5827  | mpiexec --ppn=1 -n2 ./signals-sleep | completed | 2         | nid000001,nid000002 |
+    | fea2726a-c531-4acd-b158-3fae197999dd  | 57.pbs-host  | 5827  | mpiexec --ppn=1 -n2 ./signals       | completed | 2         | nid000001,nid000002 |
+    +---------------------------------------+--------------+-------+-------------------------------------+-----------+-----------+---------------------+
+
+SEE ALSO
+========
+
+sat(8)
+
+.. include:: _notice.rst
--- a/sat/apiclient/jobstat.py
+++ b/sat/apiclient/jobstat.py
@@ -1,0 +1,60 @@
+#
+# MIT License
+#
+# (C) Copyright 2022 Hewlett Packard Enterprise Development LP
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
+# in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+"""
+Client for querying the Job State Checker service
+"""
+
+import logging
+
+from sat.apiclient.gateway import APIError, APIGatewayClient
+
+LOGGER = logging.getLogger(__name__)
+
+
+class JobstatClient(APIGatewayClient):
+    base_resource_path = 'statechecker/jobstat/'
+
+    def get_all(self):
+        """Get all results from Job State Checker.
+
+        Returns:
+            A list of dictionaries where each dictionary pertains to a
+            single job.
+
+        Raises:
+            APIError: if there is a failure querying the Job State Checker API
+                or getting the required information from the response.
+        """
+        err_prefix = 'Failed to get State Checker data'
+        try:
+            response = self.get('all').json()
+            if 'errorcode' in response and response['errorcode'] != 0:
+                raise APIError(response.get('errormessage', 'unknown error'))
+            return response['jobstat']
+        except APIError as err:
+            raise APIError(f'{err_prefix}: {err}')
+        except ValueError as err:
+            raise APIError(f'{err_prefix} due to bad JSON in response: {err}')
+        except KeyError as err:
+            raise APIError(f'{err_prefix} due to missing {err} key in response.')
--- a/sat/cli/jobstat/main.py
+++ b/sat/cli/jobstat/main.py
@@ -1,0 +1,62 @@
+# MIT License
+#
+# (C) Copyright 2022 Hewlett Packard Enterprise Development LP
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
+# in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+"""
+The main entry point for the jobstat subcommand.
+"""
+import logging
+from sat.report import Report
+from sat.config import get_config_value
+from sat.apiclient import APIError
+from sat.apiclient.jobstat import JobstatClient
+from sat.session import SATSession
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+def do_jobstat(args):
+    """Run the jobstat command with the given arguments.
+    Args:
+        args: The argparse.Namespace object containing the parsed arguments
+            passed to this subcommand.
+    """
+    state_checker_client = JobstatClient(SATSession())
+    try:
+        application_data = state_checker_client.get_all()
+        if not application_data:
+            return
+        report = Report(
+            tuple(application_data[0].keys()),
+            None,
+            args.sort_by,
+            args.reverse,
+            filter_strs=args.filter_strs,
+            display_headings=args.fields,
+            print_format=args.format,
+        )
+        for row in application_data:
+            report.add_row(row)
+        print(report)
+    except APIError as err:
+        LOGGER.error(err)
+        raise SystemExit(1)
--- a/sat/cli/jobstat/parser.py
+++ b/sat/cli/jobstat/parser.py
@@ -1,0 +1,46 @@
+#
+# MIT License
+#
+# (C) Copyright 2022 Hewlett Packard Enterprise Development LP
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
+# in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+"""
+The parser for the jobstat subcommand.
+"""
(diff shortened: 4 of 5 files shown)


docs/man/sat-jobstat.8.rst
```diff
@@ -0,0 +1,61 @@
+=============
+ SAT-JOBSTAT
+=============
+
+-----------------------------------------------------
+Jobstat - Check status of job(s) and its application(s)
+-----------------------------------------------------
```
ekamin-hpe, line 7: Match the number of characters in the underline/overline to the title.

```suggestion
-------------------------------------------------------
Jobstat - Check status of job(s) and its application(s)
-------------------------------------------------------
```

ekamin-hpe (changes requested):
There are still some unaddressed comments on the original PR:

https://github.com/Cray-HPE/sat/pull/38#discussion_r950511118
https://github.com/Cray-HPE/sat/pull/38#discussion_r950513754
https://github.com/Cray-HPE/sat/pull/38#discussion_r950516301
https://github.com/Cray-HPE/sat/pull/38#discussion_r950517035
https://github.com/Cray-HPE/sat/pull/38#discussion_r950519334
https://github.com/Cray-HPE/sat/pull/38#discussion_r950525281
https://github.com/Cray-HPE/sat/pull/38#discussion_r950525875
https://github.com/Cray-HPE/sat/pull/38#discussion_r950526388

I am also still curious if there is going to be authentication on the API endpoint as I mentioned in this comment, and we should be using the config file field for the API gateway hostname rather than hard-coding it: https://github.com/Cray-HPE/sat/pull/38#discussion_r968926035

Finally, as I brought up in that last comment, I don't really like that we are using `requests` in a different way than the other commands, which all use the `APIGatewayClient` class by deriving a subclass from it. Maybe that is something we can help with, if my last comment was not clear on how to proceed.

docs/man/sat-jobstat.8.rst
```diff
@@ -0,0 +1,61 @@
...
+
+SYNOPSIS
+========
+
+**sat** [global-opts] **jobstat** [options]
+
+DESCRIPTION
+===========
+
+Jobstat is a command that allows a user or admin to access application and job data via the command line. The command provides a table showing the user the apid, jobid, user, state, and time-reported of all jobs on the system.
```
ekamin-hpe, line 21: Sorry for not noticing this earlier, but could this line be wrapped at 80 characters?

sat/cli/jobstat/parser.py
```diff
@@ -0,0 +1,46 @@
...
+    Returns:
+        None
+    """
+    format_options = sat.parsergroups.create_format_options()
+    filter_options = sat.parsergroups.create_filter_options()
+
+    jobstat_parser = subparsers.add_parser(
+        'jobstat',
+        help='Application information',
+        description='This command shows all jobs and relevent information pertaining to those applications.',
```
ekamin-hpe, line 44: nit: "relevent" is misspelled.

sat/cli/jobstat/main.py
```diff
@@ -0,0 +1,72 @@
...
+from sat.config import get_config_value
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+def unload_json(url):
+    """Unloads the JSON object from the server url and formats the data as a dictionary"""
+    try:
+        connect = APIGatewayClient()
```
ekamin-hpe, line 38: I think this would potentially work though it won't work if the API Gateway is authenticated.

We don't usually instantiate `APIGatewayClient` directly, instead we always create subclasses for the different services.

Here is a gist explaining what I think should change: https://gist.github.hpe.com/eli-kamin/3675a843e0b343f3360f2bcd11327da1

sat/cli/jobstat/main.py
```diff
@@ -0,0 +1,72 @@
...
+            get_config_value('format.no_headings'),
+            get_config_value('format.no_borders'),
+            filter_strs=args.filter_strs,
+            display_headings=args.fields,
+            print_format=args.format,
+        )
+        for row in application_data:
+            report.add_row(row)
+        print(report)
+    except Exception as err:
```
ekamin-hpe, line 70: I would prefer that we don't catch the bare exception class here.

We should most likely only need to catch HTTP errors coming from the API. This comment is also addressed in the gist I left: https://gist.github.hpe.com/eli-kamin/3675a843e0b343f3360f2bcd11327da1

docs/man/sat-jobstat.8.rst
```diff
@@ -0,0 +1,59 @@
...
+apid, jobid, user, state, and time-reported of all jobs on the system.
+
+OPTIONS
+=======
+
+These options must be specified after the subcommand.
+
+**-h, --help**
+        Print the help message for 'sat jobstat'.
+.. include:: _sat-format-opts.rst
```
ekamin-hpe, line 32: Please add an empty line between lines 31 and 32. As it currently exists, compiling the man pages gives a warning:
```
sat-jobstat.8.rst:32: (WARNING/2) Definition list ends without a blank line; unexpected unindent.
```

ekamin-hpe (changes requested):
This is getting much closer, though I would really still like to refactor the API client code to be consistent with the other commands in SAT. I hope the gist I provided helps with that.

Additionally, some level of testing would be necessary - optimally, we would include a description of testing performed in the commit message. Is this status checker a service that will be available generally on internal systems? Is it part of some product stream?

Once we're ready to move forward with this, you'll need to squash all the commits in this pull request to one so we can keep the commit history clean, and I will help with getting it checked in.

sat/cli/jobstat/main.py
```diff
@@ -24,49 +24,40 @@
...
-            filter_strs=args.filter_strs,
-            display_headings=args.fields,
-            print_format=args.format,
-        )
-        for row in application_data:
-            report.add_row(row)
-        print(report)
-    except Exception as err:
-        LOGGER.error(f'Failed to execute jobstat command: {err}')
+        if args.all is True:
```
ekamin-hpe, line 45: Can this case be removed again? I think in my gist I did not remove it, sorry about that.

sat/apiclient/state_checker.py
```diff
@@ -0,0 +1,56 @@
...
+"""
+Client for querying the State Checker service
+"""
+
+import logging
+
+from sat.apiclient.gateway import APIError, APIGatewayClient
+
+LOGGER = logging.getLogger(__name__)
+
```
ekamin-hpe, line 33: Can you please add an extra blank line here? Otherwise we get a code style issue:
```
mbp-ekamin:sat ekamin$ pycodestyle --config pycodestyle.conf sat tests
sat/apiclient/state_checker.py:34:1: E302 expected 2 blank lines, found 1
```

sat/apiclient/state_checker.py
```diff
@@ -0,0 +1,56 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+"""
+Client for querying the State Checker service
```
ekamin-hpe, line 25: Sorry that I didn't do this in my original gist, but I am thinking it might be a good idea to rename the class `JobstatStateChecker` instead of just "state checker", otherwise it is really vague. There are so many different clients in SAT, that it might be misleading and sound like this is for HSM state or some other "state" service.

That would entail this in a few different places, specifically:
* The name of the class
* The name of this file
* Any mentions in docs or error messages (e.g. lines 38 and 48)

ekamin-hpe (commented):
A few more comments, but this is looking very good. Prior to merging I would like the branch to be rebased and have the various commits squashed into one, and if you could put a test description in the commit message that would be helpful. I am also thinking unit tests could still be helpful, though perhaps not urgently needed if manual testing can be done.

I am also going to change the target of this PR to be `feature/PE-42104-sat-jobstat`. What will need to happen once this is ready to be checked in is that we will merge it to that target branch, at which point we will verify the build is working and potentially add any last-minute fixes. We might add unit tests at this stage as well. Then we will check into integration.

sat/cli/jobstat/main.py
```diff
@@ -0,0 +1,63 @@
...
+def do_jobstat(args):
+    """Run the jobstat command with the given arguments.
+    Args:
+        args: The argparse.Namespace object containing the parsed arguments
+            passed to this subcommand.
+    """
+    state_checker_client = StateCheckerClient(SATSession())
+    try:
+        if args.all is True:
+            application_data = state_checker_client.get_all()
```
ekamin-hpe, line 46: I believe we now need to account for the case when `application_data` is an empty list, otherwise I see the following error:
```
mbp-ekamin:sat ekamin$ sat jobstat
...
Traceback (most recent call last):
  File "/Users/ekamin/.virtualenvs/sat/bin/sat", line 33, in <module>
    sys.exit(load_entry_point('sat', 'console_scripts', 'sat')())
  File "/Users/ekamin/code/sat/sat/main.py", line 120, in main
    subcommand(args)
  File "/Users/ekamin/code/sat/sat/cli/jobstat/main.py", line 47, in do_jobstat
    tuple(application_data[0].keys()),
IndexError: list index out of range
```

sat/apiclient/state_checker.py
```diff
@@ -0,0 +1,56 @@
...
+            A list of dictionaries where each dictionary pertains to a
+            single job.
+
+        Raises:
+            APIError: if there is a failure querying the State Checker API
+                or getting the required information from the response.
+        """
+        err_prefix = 'Failed to get State Checker data'
+        try:
+            return self.get('all').json()['jobstat']
```
ekamin-hpe, line 50: I am getting some weird behavior running through this code. The state checker service seems to be returning `200` code responses but are actually error cases, and as a result `APIError` is not being raised here as we would expect.

```
>>> resp = self.get('all')
>>> resp
<Response [200]>
```
But in the actual JSON there is an error:
```
>>> resp.json()
{'errorcode': 4, 'errormessage': 'Problem processing connection request', 'timereported': '2022-09-19T18:34:24', 'jobstat': []}
```

It seems like this should really be addressed in the `jobstat` service (codes of 4XX or 5XX will automatically raise an error). Maybe you could work around it here by raising an `APIError` if the returned JSON contained an `errorcode` field, but I wouldn't say that's an optimal solution.

```
        try:
            response = self.get('all').json()
            error_code = response.get('errorcode')
            if error_code is not None and error_code != 0:
                raise APIError(f'{response.get("errormessage", "error in state checker API")}')
            return response['jobstat']
        except APIError as err:
            raise APIError(f'{err_prefix}: {err}')
```

sat/cli/jobstat/main.py
```diff
@@ -42,10 +42,10 @@ def do_jobstat(args):
     """
     state_checker_client = StateCheckerClient(SATSession())
     try:
-        if args.all is True:
-            application_data = state_checker_client.get_all()
+        application_data = state_checker_client.get_all()
+        if application_data['jobstat'] == None:
```
ekamin-hpe, line 46: I am a little confused by the most recent change on this line, but I think you probably meant to do this as `is not None` rather than `== None`, otherwise you are only creating the report if the response looks like this:
```
{"jobstat": null}
```
(i.e. it contained a `'jobstat'` key and the value was `NoneType`).

With this change, I also don't think it's going to be correctly handling the case when the response is missing a `jobstat` key, the command will fail with an uncaught `KeyError` coming from this line.

Assuming you did mean to say `if application_data['jobstat'] is not None`, then If the response does contain a `jobstat` key but returns an empty list, I think you will get an uncaught `IndexError` on line 48 trying to get the 0th element of an empty list.

I think it would be good to also account for the case when the response contains the `jobstat` key but a `null` value (i.e. that case I show at the top of this comment), once again I think that would result in an unhandled exception coming from line 48.

Just in general I think it would still be cleaner to be returning just the value pointed to by the `"jobstat"` key - this version makes the `except KeyError` on line 55 of `state_checker.py` no longer necessary - and I don't think we are correctly handling that case now.

For me I think I would still prefer something along these lines, though there might be cases that I am not thinking of: https://gist.github.hpe.com/eli-kamin/aa8f60f88dd465507f6410ee568e98ba

sat/cli/jobstat/main.py
```diff
@@ -55,9 +55,12 @@ def do_jobstat(args):
                 display_headings=args.fields,
                 print_format=args.format,
             )
-            for row in application_data:
+            for row in application_data['jobstat']:
                 report.add_row(row)
             print(report)
+        else:
+            # Print error message if response did not have jobstat's
+            print(f"jobstat error: {application_data['errormessage']}")
```
ekamin-hpe, line 63: I would much prefer if we could keep this error handling in the `StateCheckerClient` class in the form of raising an `APIError` (see other comment), but if we must do it here, let's use `LOGGER.error` instead of `print`, so that the error goes to `stderr`.

ekamin-hpe (changes requested):
I have some comments regarding error handling.

sat/cli/jobstat/main.py
ekamin-hpe: Are we attempting to handle the case when the API returns both valid jobstat results as well as an error message? It seems like maybe you are trying to do that, not sure.

vinaybabug-hpe: Our response always has jobstat e.g. 
`{'machine': 'slice', 'errorcode': 2, 'errormessage': 'Unable to query Elasticsearch', 'timereported': '2022-09-20T08:26:02', 'jobstat': []}`
We have machine just in case for multi tenancy.  A non-zero errorcode indicates failure so we use that instead of suggested checks.

vinaybabug-hpe: Response structure is always the same from the server, just values change.

sat/apiclient/JobstatStateChecker.py
```diff
@@ -0,0 +1,60 @@
...
+
+        Raises:
+            APIError: if there is a failure querying the Job State Checker API
+                or getting the required information from the response.
+        """
+        err_prefix = 'Failed to get State Checker data'
+        try:
+            response = self.get('all').json()
+            if response['errorcode'] != 0:
+                raise APIError(response['errormessage'])
```
ekamin-hpe, line 53: This is pretty close to what I was hoping for, thanks for addressing all the feedback here.

I see you mentioned that the response objects always have `errorcode` and `errormessage` keys (even in the successful case) but I think it would be good to be more defensive and handle when they aren't there, i.e. first check that the expected keys exist.

Otherwise, in the case when there is successful `jobstat` data but there is no errorcode, we will fail with an `APIError` (due to missing key). Other (edge cases) thoughts I have are how we are handling if `errorcode` is a null type, and how we are handling if the 'errormessage' key is missing.

Just in case the behavior of the API changes, or something unexpected happens.

How would you feel about this approach, I think it's a bit more defensive:
```suggestion
            if 'errorcode' in response and response['errorcode']:
                raise APIError(response.get('errormessage', 'unknown error'))
```

ekamin-hpe (commented):
I have one final comment, I think after that I would be ready to approve. There seems to also be a conflict in the change log that needs resolving.

When you're done pushing changes, could you please squash all your commits into one and include a short description of the testing performed? I personally was unable to test this command in the successful case, every time I tried running it I only got errors from the service.

sat/cli/jobstat/main.py
```diff
@@ -0,0 +1,64 @@
...
+        application_data = state_checker_client.get_all()
+        if not application_data:
+            return
+        report = Report(
+            tuple(application_data[0].keys()),
+            None,
+            args.sort_by,
+            args.reverse,
+            get_config_value("format.no_headings"),
+            get_config_value("format.no_borders"),
```
jack-stanek-hpe, line 54: FWIW, these arguments aren't needed. When the `no_headings` and `no_borders` arguments are `None` (the default), the `Report` class looks up the config file values. See here: https://github.com/vinaybabug-hpe/sat/blob/993cf683b89798958244ac2ab4473f5e8650bef0/sat/report.py#L102-L114

jack-stanek-hpe (commented):
Where can I find a specification for the jobstat API (e.g. OpenAPI or Swagger specs?)

sat/apiclient/JobstatStateChecker.py
```diff
@@ -0,0 +1,60 @@
...
+"""
+
+import logging
+
+from sat.apiclient.gateway import APIError, APIGatewayClient
+
+LOGGER = logging.getLogger(__name__)
+
+
+class JobstatStateChecker(APIGatewayClient):
```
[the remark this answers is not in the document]
vinaybabug-hpe: For changes regarding APIError; I will defer to sat team as the current codebase does not mention changes.

sat/cli/jobstat/main.py
vinaybabug-hpe: removed them

sat/apiclient/jobstat.py
```diff
@@ -0,0 +1,62 @@
...
+
+        Raises:
+            APIError: if there is a failure querying the Job State Checker API
+                or getting the required information from the response.
+        """
+        err_prefix = 'Failed to get State Checker data'
+        try:
+            response = self.get('all').json()
+            if 'errorcode' in response and response['errorcode'] != 0:
+                raise APIError(response['errormessage'])
```
ekamin-hpe, line 53: I would slightly prefer `APIError(response.get('errormessage'))` so that we don't raise a `KeyError` if `errormessage` is unexpectedly missing. Possibly, `APIError(response.get('errormessage', 'unknown error'))`.

sat/apiclient/jobstat.py
```diff
@@ -0,0 +1,62 @@
...
+            if 'jobstat' in response:
```
ekamin-hpe, line 54: I would actually say we don't necessarily need to check `if 'jobstat' in response` here, because this would be handled by the `except KeyError` on line 61. If the `jobstat` key is missing and we didn't get any error reported, I think we should be raising an `APIError`.

sat/apiclient/jobstat.py
```diff
@@ -0,0 +1,62 @@
...
+                return response['jobstat']
+            return None
```
ekamin-hpe, line 56: This line is not necessary because functions will return `None` unless otherwise specified. If we want a default case, I'd make it an empty list, but I would prefer if the `jobstat` key is missing to be raising an error (as I mentioned in the comment above).

The reason I am thinking of the `errorcode` and `jobstat` fields as different in terms of how we should be checking them is that in the case that the `errorcode` field is missing but `jobstat` key is not, I think we should be able to return the jobstat data without an error. But in the cases:
* when both the `jobstat` key and the `errorcode` key are missing
* when the `errorcode` was 0 but the `jobstat` key was still missing

These are cases when I think we should be raising an error.

docs/man/sat-jobstat.8.rst
```diff
@@ -0,0 +1,60 @@
+=============
+ SAT-JOBSTAT
+=============
+
+-------------------------------------------------------
```
ekamin-hpe, line 5: nit: now that line 6 has been shortened could you shorten these lines to match as well? (This line and line 7).

ekamin-hpe (commented):
Sorry to keep nitpicking the implementation of `JobstatClient.get_all`, I appreciate your continuing to refine it. I have a few more comments about this version. Here is what I was thinking for this function:

https://gist.github.hpe.com/eli-kamin/59d01cad892f88e5a73bc2ce26ab68ed

sat/apiclient/jobstat.py
vinaybabug-hpe: well, missing errorcode/errormessage would have resulted in the same. This just handles the case more gracefully...

sat/apiclient/jobstat.py
vinaybabug-hpe: I don't see this happening with the current server implementation. Only memory corruption could result in such scenarios.

vinaybabug-hpe:
I am not able to sync with the main sat codebase as this code is a fork from Elmir's code/fork. Can we take this to a branch in Cray-HPE/sat (not master)?

sat/apiclient/jobstat.py
ekamin-hpe: My worry is that the server implementation could change. I don't think there's ever really a good case for returning `None` from this function, I think it should either return valid results or raise an error.

sat/apiclient/jobstat.py
ekamin-hpe: I was thinking of the case when the server returned valid jobstat results but for some reason didn't return errorcode/errormessage keys. I think in that case we would not want to raise an error.

sat/apiclient/JobstatStateChecker.py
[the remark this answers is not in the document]
ekamin-hpe: Depending on the timing, this may be something that the SAT team will need to address after this has been merged.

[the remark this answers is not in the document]
jack-stanek-hpe: Correct, no changes are needed at this point.

ekamin-hpe (commented):
Just a quick comment on the commit message, "Refactored code according to comments by using PIGatewayClient" is missing a letter, should be "Refactored code according to comments by using APIGatewayClient"

ekamin-hpe (approved):
Looks good, thank you again for the sustained effort to address all the feedback. This will be merged to the `feature/PE-42104-sat-jobstat` branch in the main Cray-HPE/sat repository, at which point we will take it over and ensure it passes all the necessary checks to go into integration.

PR #59, opened 2022-09-16
CRAYSAT-1569: Use `csm-api-client` library

jack-stanek-hpe:
## Summary and Scope

This PR adds a dependency on the newly factored-out csm-api-client library,
and removes the API client classes which were factored out. The
remaining client classes were made subclasses of the common library's
API client base classes. Additional refactors were made to accomodate
the new library's API, as SAT-specific code was moved out of the API
client classes.

This PR also makes use of the UserSession class from the csm-api-client
library to maintain compatibility with the old functionality of the
SATSession class. The new SATSession class only handles authentication
and SAT-specific handling of the configuration file.

## Issues and Related PRs

* Resolves [CRAYSAT-1569](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1569)

### Tested on:

  * `baldar`

### Test description:
Run unit tests. Test simple `sat` commands on local machine against remote API endpoint. Test authenticating against remote API endpoint.

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated


jack-stanek-hpe:
Note that builds for this change will fail until the PR is merged on the [`python-csm-api-client`](https://github.com/Cray-HPE/python-csm-api-client/pull/1) library which creates the 1.0.0 release.

ekamin-hpe (approved):
Looks good to me.

CHANGELOG.md
```diff
@@ -29,6 +29,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Updated the README file to remove outdated information.
+- Moved API client classes to `csm-api-client` library, and added that new
```
haasken-hpe, line 32: Just a reminder to rebase and move this to a new "Unreleased" section.

haasken-hpe, line 32: Suggest clarifying a bit because not all were moved, just `APIGatewayClient`, `CFSClient`, `HSMClient`, `VCSRepo`, and associated error classes and the `SATSession` functionality was refactored mostly into `UserSession`. Perhaps just "Refactored to use common `APIGatewayClient`, `CFSClient`, `HSMClient`, `VCSClient`, and `UserSession` classes from `csm-api-client` library."

sat/apiclient/fas.py
```diff
@@ -24,14 +24,14 @@
 """
 Client for querying the Firmware Action Service (FAS) API.
 """
+from datetime import datetime, timedelta
 import json
 import logging
 import time
-from datetime import datetime, timedelta
 
+from csm_api_client.service.gateway import APIError, APIGatewayClient
```
haasken-hpe, line 32: There appears to be a type-checking error worth looking into on line 350 of this file. Here's the relevant snippet of code:

```python
        # create a system snapshot set to expire in 10 minutes
        payload = {'name': name, 'expirationTime': later}

        # filter on xnames if provided.
        if xnames:
            payload['stateComponentFilter'] = {'xnames': xnames}

        payload = json.dumps(payload, cls=_DateTimeEncoder)

        try:
            self.post('snapshots', payload=payload)
        except APIError as err:
            raise APIError('Error when posting new snapshot: {}'.format(err))
```

The error I'm seeing in PyCharm says, "Expected type 'Optional[dict]', got 'str' instead". Can you look into this and see if it's going to be a problem? Perhaps test `sat firmware` to double check.

sat/apiclient/telemetry.py
```diff
@@ -26,7 +26,7 @@
 """
 import logging
 
-from sat.apiclient.gateway import APIError, APIGatewayClient, ReadTimeout
+from csm_api_client.service.gateway import APIError, APIGatewayClient, ReadTimeout
```
haasken-hpe, line 29: PyCharm doesn't like the signature of the `TelemetryAPIClient.stream` method, saying "Signature of method 'TelemetryAPIClient.stream()' does not match signature of the base method in class 'APIGatewayClient'". However, this same warning exists prior to your change to use the common library, so perhaps not an issue.

haasken-hpe (approved):
All comments are pretty minor. One new type-checking error in `fas.py` that might be worth investigating a little further.

haasken-hpe:
Minor additional comment: there are many places in docstrings that specify `sat.apiclient.CFSClient` as the type of an argument. Although that still kinda works because of the import in `sat/apiclient/__init__.py`, we might want to consider updating those.

sat/apiclient/fas.py
jack-stanek-hpe: This is partly an issue with how this was originally written and how I changed the type annotations on `APIGatewayClient.post()` in `csm-api-client`. The `payload` argument is passed to the `data` argument of  `requests.request()`, which according to the [requests docs](https://requests.readthedocs.io/en/latest/api/#requests.request) can be:
> Dictionary, list of tuples, bytes, or file-like object to send in the body of the [Request](https://requests.readthedocs.io/en/latest/api/#requests.Request).

This corresponds to the type signature `Optional[Union[Dict, List[Tuple[...]], bytes, IO]]`. However, `APIGatewayClient.post()` only accepts `Optional[Dict]`. This restriction is not strictly necessary, but it corresponds to how it is being used already in SAT. It is fortunately fairly easy to work around in the caller by performing type conversions on the dict values before passing in the dict to the `post()` method.

sat/apiclient/telemetry.py
jack-stanek-hpe: Yeah, this is true. For reference, the `APIGatewayClient.stream()` method has the following type:
```python
def stream(self, *args: str, params: Optional[Dict] = None, **kwargs: Any) -> Response: ...
```
The `TelemetryAPIClient.stream()` method has the following incompatible type:
```python
def stream(self, topic: str, timeout: int, params: Optional[Dict] = None) -> Response: ...
```
As you note, this was the case before the refactor as well, though I don't think it has a practical effect in how we use the telemetry client class. (It _does_ violate the [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle) though, so strictly speaking it's not ideal.)

haasken-hpe (approved):

PR #61, opened 2022-09-19
CRAYSAT-1539: Use raise_not_ok API in subcommand implementations

jack-stanek-hpe:
## Summary and Scope

This change uses the `raise_not_ok` option in implementations of `sat swap blade` and the validators for `sat bootprep`. This allows easier changes to behavior based on HTTP status codes.

## Issues and Related PRs

* Resolves [CRAYSAT-1539](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1539)

### Tested on:

  * `mug`

### Test description:

Run unit tests. Manually perform regression test for `sat bootprep`.

## Risks and Mitigations

_Are there known issues with these changes? Any other special considerations?_

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated

### DIFF
--- a/sat/cli/bootprep/input/session_template.py
+++ b/sat/cli/bootprep/input/session_template.py
@@ -121,11 +121,16 @@ class InputSessionTemplate(BaseInputItem):
             return
 
         try:
-            self.cfs_client.get_configuration(self.configuration)
-        except APIError as err:
-            # TODO: should probably differentiate between 404 Not Found and other errors
-            raise InputItemValidateError(f'Configuration {self.configuration} specified for '
-                                         f'{self} does not exist: {err}')
+            resp = self.cfs_client.get('configurations', self.configuration, raise_not_ok=False)
+            if not resp.ok:
+                if resp.status_code == 404:
+                    raise InputItemValidateError(f'Configuration {self.configuration} specified for '
+                                                 f'{self} does not exist.')
+                else:
+                    self.cfs_client.raise_from_response(resp)
+        except (APIError, ValueError) as err:
+            raise InputItemValidateError(f'An error occurred while querying configuration '
+                                         f'{self.configuration}: {err}')
 
     @abstractmethod
     def validate_image_exists(self, **_):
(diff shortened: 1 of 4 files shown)


sat/cli/bootprep/input/session_template.py
```diff
@@ -121,11 +121,16 @@ def validate_configuration_exists(self, **_):
...
 
         try:
-            self.cfs_client.get_configuration(self.configuration)
+            resp = self.cfs_client.get('configurations', self.configuration, raise_not_ok=False)
+            if not resp.ok:
+                if resp.status_code == 404:
+                    raise InputItemValidateError(f'Configuration {self.configuration} specified for '
+                                                 f'{self} does not exist.')
+                else:
+                    self.cfs_client.raise_from_response(resp)
```
haasken-hpe, line 130: I don't think anything will catch this, will it? That will result in a traceback on an `APIError` rather than a nice validation error to the user. I would suggest that we still raise an `InputItemValidateError` in this case, but just with an error message that says that the query to CFS failed for some reason.

sat/cli/swap/blade.py
```diff
@@ -248,18 +248,15 @@ def condition_name(self):
...
-            # RedfishEndpoints so if an invalid xname is given it will time out
-            # instead of giving a proper error
-            if not any('404' in arg for arg in err.args):
-                raise WaitingFailure(f'Could not query Redfish endpoint for {member}: {err}')
+        resp = self.hsm_client.get('Inventory', 'RedfishEndpoints', member, raise_not_ok=False)
+        # Missing endpoints may have not been discovered yet, in which case
+        # querying their xnames will simply return a 404.
+        if not resp.ok and resp.status_code != 404:
+            raise WaitingFailure(f'Could not query Redfish endpoint for {member}: {resp.reason}')
+        endpoint = resp.json()
```
haasken-hpe, line 256: This new logic feels a little flawed to me. If the response was a 404, it's going to get the `resp.json()` now. In the old logic, it would have just returned without doing anything, giving it another opportunity to check on the next cycle

sat/session.py
```diff
@@ -25,145 +25,42 @@
...
 LOGGER = logging.getLogger(__name__)
 
 
-class SATSession:
-    """Manage API sessions, authentication, and token storage/retrieval."""
-
-    TOKEN_URI = '/keycloak/realms/{}/protocol/openid-connect/token'
-    tenant = 'shasta'
-    client_id = 'shasta'
+class SATSession(UserSession):
```
haasken-hpe, line 39: Why are the changes in this file showing up here in addition to in #59? It seems like they belong just in #59, unless I'm missing something.

sat/cli/bootprep/input/session_template.py
```diff
@@ -121,11 +121,16 @@ def validate_configuration_exists(self, **_):
...
         except APIError as err:
-            # TODO: should probably differentiate between 404 Not Found and other errors
```
haasken-hpe, line 126: There are still two more instances of this TODO comment in this file on lines 233 and 308 where we're checking if the image specified in a session template exists.

haasken-hpe (changes requested):
Found a few things that need to be addressed before merging.

sat/cli/bootprep/input/session_template.py
jack-stanek-hpe: This will be caught on line 131 and be re-raised as an `InputItemValidateError` below. This should behave the same way as if a non-404 error was raised previously.

sat/session.py
jack-stanek-hpe: I am unsure -- I will try rebasing against the new `CRAYSAT-1569-use-api-client-library` branch content to see if that fixes things.

sat/cli/swap/blade.py
jack-stanek-hpe: I think you're right. Something like the following should capture the behavior better:
```python
if not resp.ok:
    if resp.status_code != 404:
        raise WaitingFailure(f'Could not query Redfish endpoint for {member}: {resp.reason}')
    else:
        return False

endpoint = resp.json()
if not endpoint:
    raise WaitingFailure(f'No Redfish endpoint found for {member}')
return endpoint['DiscoveryInfo']['LastDiscoveryStatus'] == 'DiscoverOK'

```

haasken-hpe (approved):

PR #75, opened 2022-11-07
PE-43843,PE-43844: Reduce sat jobstat table width

vinaybabug-hpe:
## Summary and Scope

_Summarize what has changed. Explain why this PR is necessary. What is impacted? Is this a new feature, critical bug fix, etc?_

- Remove full path for mpiexec or aprun
- show only first 8 chars for apid
- 
_Is this change backwards incompatible, backwards compatible, or a backwards compatible bugfix?_

## Issues and Related PRs

* Resolves [issue id](issue link)
* Change will also be needed in `<insert branch name here>`
* Future work required by [issue id](issue link)
* Documentation changes required in [issue id](issue link)
* Merge with/before/after `<insert PR URL here>`

## Testing

### Tested on:

  * `Slice`

### Test description:

_How were the changes tested and success verified? If schema changes were part of this change, how were those handled in your upgrade/downgrade testing?_

- Were the install/upgrade-based validation checks/tests run (goss tests/install-validation doc)?
- Were continuous integration tests run? If not, why?
- Was upgrade tested? If not, why?
- Was downgrade tested? If not, why?
- Were new tests (or test issues/Jiras) created for this change?

## Risks and Mitigations

_Are there known issues with these changes? Any other special considerations?_

## Pull Request Checklist

- [ ] Version number(s) incremented, if applicable
- [ ] Copyrights updated
- [ ] License file intact
- [ ] Target branch correct
- [ ] CHANGELOG.md updated
- [ ] [HPC Product Announcement](https://cray.slack.com/archives/C026TVCSXLH) prepared, if applicable

### DIFF
--- a/sat/cli/jobstat/main.py
+++ b/sat/cli/jobstat/main.py
@@ -24,9 +24,11 @@
 The main entry point for the jobstat subcommand.
 """
 import logging
+import re
 import textwrap
 
-from sat.apiclient import APIError
+from csm_api_client.service.gateway import APIError
+
 from sat.apiclient.jobstat import JobstatClient
 from sat.report import Report
 from sat.session import SATSession
@@ -44,15 +46,20 @@ def wrap_long_row_lines(row):
         None. Modifies row in place.
     """
     if row.get('apid'):
-        row['apid'] = textwrap.fill(text=row['apid'], width=20)
+        row['apid'] = row['apid'][:8]
     if row.get('jobid'):
         row['jobid'] = textwrap.fill(text=row['jobid'], width=20)
     if row.get('command'):
-        row['command'] = textwrap.fill(text=row['command'], width=40)
+        tempstr = textwrap.fill(text=row['command'], width=40)
+        tempstr = re.sub(r'\A/.*/mpiexec', 'mpiexec', tempstr, 1)
+        tempstr = re.sub(r'\A/.*/aprun', 'aprun', tempstr, 1)
+        row['command'] = tempstr
     if row.get('node_list'):
         # number of nodes to print on each line
-        length = 2
-        nodelist = row['node_list'].split(',')
+        length = 4
+        nodelist = re.sub(r'nid', '', row['node_list'])
+        nodelist = nodelist.split(',')
+        nodelist = [nid.lstrip('0') for nid in nodelist]
         row['node_list'] = '\n'.join(
             ','.join(e)
             for e in [nodelist[i: i + length] for i in range(0, len(nodelist), length)]
--- a/sat/apiclient/jobstat.py
+++ b/sat/apiclient/jobstat.py
@@ -27,7 +27,7 @@ Client for querying the Job State Checker service
 
 import logging
 
-from sat.apiclient.gateway import APIError, APIGatewayClient
+from csm_api_client.service.gateway import APIError, APIGatewayClient
 
 LOGGER = logging.getLogger(__name__)
 


ekamin-hpe (changes requested):
Let's work offline to figure out the branch history issues here. I'm expecting to see a single commit based off of `feature/PE-42104-sat-jobstat` and instead am seeing 18 commits and 51 files changed.

CHANGELOG.md
```diff
@@ -32,6 +32,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   status data.
 - Added a `sat bootprep list-vars` subcommand which lists the variables
   available in bootprep input files at runtime.
+- Added new command `jobstat`, which provides system-wide view of application
```
ekamin-hpe, line 35: let's remove this - this change log entry is already on line 31

sat/cli/jobstat/main.py
```diff
@@ -25,8 +25,9 @@
 """
 import logging
 import textwrap
+import re
```
ekamin-hpe, line 28: nit: alphabetize imports
```suggestion
import logging
import re
import textwrap
```

sat/cli/jobstat/main.py
```diff
@@ -44,11 +45,16 @@ def wrap_long_row_lines(row):
         None. Modifies row in place.
     """
     if row.get('apid'):
-        row['apid'] = textwrap.fill(text=row['apid'], width=20)
+        row['apid'] = row['apid'][:8]+"..."
```
ekamin-hpe, line 48: nit: try to use single quotes more consistently. I also would prefer spaces around the `+` operator as a small style suggestion.
```suggestion
        row['apid'] = row['apid'][:8] + '...'
```

sat/cli/jobstat/main.py
```diff
@@ -44,11 +45,16 @@ def wrap_long_row_lines(row):
...
     if row.get('jobid'):
         row['jobid'] = textwrap.fill(text=row['jobid'], width=20)
     if row.get('command'):
-        row['command'] = textwrap.fill(text=row['command'], width=40)
+        tempstr = textwrap.fill(text=row['command'], width=40)
+        if re.findall("\A/.*/mpiexec", tempstr):
+            tempstr = re.sub("\A/.*/mpiexec", "mpiexec", tempstr, 1)
```
ekamin-hpe, line 54: I get `pycodestyle` violations from these regexes:
```
mbp-ekamin:sat ekamin$ pycodestyle --config pycodestyle.conf sat tests
sat/cli/jobstat/main.py:53:24: W605 invalid escape sequence '\A'
sat/cli/jobstat/main.py:54:31: W605 invalid escape sequence '\A'
sat/cli/jobstat/main.py:55:26: W605 invalid escape sequence '\A'
sat/cli/jobstat/main.py:56:31: W605 invalid escape sequence '\A'
```

If you specify as raw strings that will fix the warnings, e.g. using `r'\A/.*/mpiexec...'`

Also, like before, would prefer that we use single quotes to stay consistent.

Getting a little nitpicky but also might suggest that we use `re.match` or `re.search` rather than `re.findall` - `re.findall` returns a list of matches and we don't actually do anything with those, I think what we're trying to do is just see if the start of the string matches the regex so something that returns just a single `match` object might be better.

Note that with `re.match` we don't strictly need the `'\A'` because `re.match` always matches at the start of a string, but if you wanted to use the same regex for both ll 53 and 54 it's probably fine to keep the `\A` character.

Overall I would probably write these lines as:

```suggestion
        mpiexec_command_regex = r'\A/.*/mpiexec'
        if re.match(mpiexec_command_regex, tempstr):
            tempstr = re.sub(mpiexec_command_regex, 'mpiexec', tempstr, 1)
```

sat/cli/jobstat/main.py
```diff
@@ -25,8 +25,9 @@
 """
 import logging
 import textwrap
+import re
 
-from sat.apiclient import APIError
+from csm_api_client.service.gateway import APIError
 from sat.apiclient.jobstat import JobstatClient
```
ekamin-hpe, line 31: Micro nit: would prefer a blank line between these two imports. We tend to group our imports as follows:
```
python std lib imports
<empty line>
third-party library imports
<empty line>
imports from the sat package
```

In this case `csm_api_client` is closest to being a third-party package (while it is authored by HPE, it's external to `sat`).

```suggestion
from csm_api_client.service.gateway import APIError

from sat.apiclient.jobstat import 
```

sat/cli/jobstat/main.py
ekamin-hpe: I am also wondering if we even need to be checking with `re.search`/`re.findall`/`re.match`. Won't `re.sub` just return the original string if the regex is not matched? Why not just this, with no `if` cases?
```
tempstr = re.sub(r'\A/.*/mpiexec', 'mpiexec', tempstr, 1)
tempstr = re.sub(r'\A/.*/aprun', 'aprun', tempstr, 1)
```

ekamin-hpe (changes requested):
Overall this looks good, though I have a few comments mostly related to code style that I'd like to get fixed up before merging.

sat/cli/jobstat/main.py
```diff
@@ -44,18 +46,23 @@ def wrap_long_row_lines(row):
...
-        length = 2
-        nodelist = row['node_list'].split(',')
+        length = 4
+        nodelist = re.sub(r'nid', '', row['node_list'])
+        nodelist = nodelist.split(',')
+        nodelist = [nid.lstrip('0') for nid in nodelist]
         row['node_list'] = '\n'.join(
             ','.join(e)
-            for e in [nodelist[i: i + length] for i in range(0, len(nodelist), length)]
+            for e in [nodelist[i : i + length] for i in range(0, len(nodelist), length)]
```
ekamin-hpe, line 65: The spacing after the `:` character results in a style violation:
```
mbp-ekamin:sat ekamin$ pycodestyle --config pycodestyle.conf sat tests
sat/cli/jobstat/main.py:65:33: E203 whitespace before ':'
```

sat/cli/jobstat/main.py
```diff
@@ -44,18 +46,23 @@ def wrap_long_row_lines(row):
         None. Modifies row in place.
     """
     if row.get('apid'):
-        row['apid'] = textwrap.fill(text=row['apid'], width=20)
+        row['apid'] = row['apid'][:8]
```
ekamin-hpe, line 49: Just checking, was it intentional to remove the trailing "..." now? My suggestion was just to change the spacing around the `+` operator, but I'm fine with this as well.

ekamin-hpe (changes requested):
Let's fix the new style violation, then this should be good to go.

sat/cli/jobstat/main.py
vinaybabug-hpe: this was recommended by our team; as first 8 chars is good enough.

ekamin-hpe (approved):

PR #85, opened 2023-01-10
CRAYSAT-1606: Implement `--limit` option to `sat bootprep run`

haasken-hpe:
## Summary and Scope

Implement a `--limit` option that allows the user to limit which types
of items from the input file are created. For example, the admin can
choose to create only the CFS configurations or only the IMS images and
BOS session templates.

This functionality is needed by the IUF because the stages are split
into `update_cfs_config` which will create the CFS configurations and
`prepare_images` which will create and configure IMS images and create
BOS session templates.

Also created constants for the keys in the input file and used it when
reading those properties from the input file and as the possible values
for the `--limit` option.

## Issues and Related PRs

* Resolves [CRAYSAT-1606](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1606)

## Testing

Not tested yet.

### Tested on:

  * frigg

## Risks and Mitigations

New feature being added to some complicated code. Risks will be mitigated by manual testing of various edge cases on frigg.

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated

### DIFF
--- a/sat/cli/bootprep/image.py
+++ b/sat/cli/bootprep/image.py
@@ -28,11 +28,11 @@ from collections import defaultdict, Counter
 import logging
 import math
 
-from sat.apiclient import APIError, CFSClient, IMSClient
+from sat.apiclient import APIError
+from sat.cli.bootprep.constants import CONFIGURATIONS_KEY, IMAGES_KEY
 from sat.cli.bootprep.errors import ImageCreateError
 from sat.cli.bootprep.input.image import DependentInputImage, IMSInputImage
 from sat.cli.bootprep.public_key import get_ims_public_key_id
-from sat.session import SATSession
 from sat.util import pester_choices
 from sat.waiting import (
     DependencyCycleError,
@@ -324,7 +324,7 @@ def validate_image_ims_bases(input_images):
                                f'input images')
 
 
-def validate_image_configurations(input_images, cfs_client, input_config_names, dry_run):
+def validate_image_configurations(input_images, cfs_client, input_config_names, input_configs_exist):
     """Validate that the images refer to valid existing configurations.
 
     Args:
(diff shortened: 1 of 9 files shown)


sat/cli/bootprep/image.py
```diff
@@ -451,18 +444,53 @@ def create_images(instance, args):
...
+        # assume the configurations from the input file do exist.
+        input_configs_exist = CONFIGURATIONS_KEY in args.limit and args.dry_run
+        # Raises ImageCreateError if validation of CFS configurations fails
+        validate_image_configurations(input_images, cfs_client,
+                                      instance.input_configuration_names, input_configs_exist)
 
     # This is where name properties are first accessed and thus rendered
     validate_unique_image_names(input_images)
 
+    # TODO: Note the slight change in behavior here. Test this use case.
```
ekamin-hpe, line 459: If I'm understanding this correctly, I'm not too worried about the change in behavior here. We're checking to make sure that an image in the config isn't accidentally the base of another image, and previously we would have allowed the case in which the conflicting image was skipped? I can't think of a case in which this couldn't be resolved just by setting up the ref_name / image_ref properties correctly. Even if there are cases that worked previously that will start failing validation now, they probably aren't what the user intended anyway.

ekamin-hpe (approved):
This looks good to me, assuming you've had time to test it.

sat/cli/bootprep/image.py
haasken-hpe: That's right. Yeah, that's what I was thinking. I would still like to try to test this use case to make sure it doesn't blow up. I meant to do that, but I forgot. I will then probably remove this TODO comment.

jack-stanek-hpe (approved):
This looks fine to me, assuming the testing went well.

ekamin-hpe (approved):
Conflict resolution looked good to me.

sat/cli/bootprep/image.py
haasken-hpe: Here are the results of a test of this use case, showing before and after behavior for this case: https://gist.github.com/haasken-hpe/7cde42a46cb7a8c8766f88db902f940e

I think you could make a case for either behavior being correct, and it's a pretty specific edge case, so I think it's fine the way it is now with this change. I'll remove the comment.

haasken-hpe:
Testing results here: https://gist.github.com/haasken-hpe/c1e696bca68af89a1d07b2a239c2cae0

Note I tested before the commit to remove the TODO comment that Eli commented on and before the rebase that squashed that commit. That's all that should have changed though.

PR #94, opened 2023-01-27
CRAYSAT-1659: Add ability to filter product images/recipes by prefix

haasken-hpe:
## Summary and Scope
If a product provides more than one image or recipe, it was previously impossible to specify any image or recipe provided by that product. Now it is possible for the admin to differentiate which image they intended by specifying a prefix which will be matched against the name of the image or recipes delivered by that product.

This is needed to support selecting between the Kubernetes image for masters and workers and the storage image for storage nodes, both of which are delivered by CSM and can be differentiated by a prefix.
## Issues and Related PRs

* Resolves [CRAYSAT-1659](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1659)
* Still needs to be documented in docs-sat

### Tested on:

  * frigg (to-do)

### Test description:
Added unit tests. Will test on frigg with manually populated product catalog data for the CSM product.

## Risks and Mitigations

Pretty low risk.

## Pull Request Checklist

- [ ] Version number(s) incremented, if applicable
- [ ] Copyrights updated
- [ ] License file intact
- [ ] Target branch correct
- [ ] CHANGELOG.md updated
- [ ] [HPC Product Announcement](https://cray.slack.com/archives/C026TVCSXLH) prepared, if applicable

### DIFF
--- a/sat/cli/bootprep/input/image.py
+++ b/sat/cli/bootprep/input/image.py
@@ -38,6 +38,7 @@ from sat.cached_property import cached_property
 from sat.cli.bootprep.errors import ImageCreateError
 from sat.cli.bootprep.input.base import jinja_rendered, provides_context
 from sat.constants import MISSING_VALUE
+from sat.util import get_val_by_path
 from sat.waiting import (
     DependencyGroupMember,
     WaitingFailure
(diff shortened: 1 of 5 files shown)


haasken-hpe:
The force push was just updating copyright years to satisfy the license check.

sat/cli/bootprep/input/image.py
```diff
@@ -639,14 +639,42 @@ def product_version(self):
         return self.image_data['base']['product'].get('version')
 
     @property
-    def base_description(self):
-        """str: a human-readable description of the base we are starting from"""
+    def filter_prefix(self):
+        """str or None: the filter prefix to use for filtering images or recipes from the product"""
+        return self.image_data['base']['product'].get('filter', {}).get('prefix')
```
jack-stanek-hpe, line 644: nit: could this be replaced with the following?
```suggestion
        return get_val_by_path(self.image_data,  'base.product.filter.prefix')
```

sat/cli/bootprep/input/image.py
```diff
@@ -639,14 +639,42 @@ def product_version(self):
...
+
+    @property
+    def filter_func(self):
+        """Callable: a function that returns True if the given name matches the filter, False otherwise"""
+        if not self.filter_prefix:
+            return lambda x: True
+        else:
+            return lambda x: x.startswith(self.filter_prefix)
```
jack-stanek-hpe, line 652: I'm a little unsure why this needs to be a property returning a function, could it be simplified with the following?
```suggestion
    def filter_func(self, name):
        """Check if the name matches the filter"""
        if not self.filter_prefix:
            return True
        return x.startswith(self.filter_prefix)
```
At callsites of the higher-order function, the usage _should_ be exactly the same, but there's one less level of abstraction involved.

ekamin-hpe (approved):
Looks good to me!

jack-stanek-hpe (approved):
Looks good, just had a couple comments for small simplifications of the implementation.

sat/cli/bootprep/input/image.py
haasken-hpe: Yeah, not sure how I forgot about that useful function.

sat/cli/bootprep/input/image.py
haasken-hpe: Yep, looks good. I guess it was just late when I implemented this 🤷

haasken-hpe: I've also added a test for when the filter prefix is the empty string, which I believe this was handling incorrectly before, and now it is handling correctly (empty string prefix matches all).

haasken-hpe: Actually, it was handling it fine since `name.startswith(self.filter_prefix)` evaluates to True just as `not self.filter_prefix` when `self.filter_prefix` is the empty string.

haasken-hpe:
Testing on frigg looked good apart from a failure of IMS to rename images (with 500 error). I've seen this issue before, and it's not related to this change. Mostly looking to see that the image bases are found and errors are handled correctly if not with this change.

Full results here: https://gist.github.com/haasken-hpe/904ca435732c75bf9e1909fb5d234b8a

haasken-hpe:
Corresponding docs-sat changes will be tracked under [CRAYSAT-1663](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1663)

PR #95, opened 2023-01-27
CRAYSAT-1653: Replace hyphens with underscores in top-level keys

ekamin-hpe:
This commit changes how bootprep variables are processed. This applies to variables defined in the software recipe, variables provided on the command line, and variables in local files (i.e. passed in with --vars-file).

This changes top-level keys of dictionaries to replace hyphens with underscores. This gives us the ability to support product names that contain hyphens. To refer to a product name that contains hyphens, the bootprep input file will need to refer to that product with its hyphens replaced with underscores.

Test Description: I ran sat bootprep using an input file that creates a CFS configuration. Then, using `--save-files` I inspected the resulting configuration that was created locally.

With a vars file containing a product with a hyphenated name:

    sat bootprep run --dry-run --vars-file vars.yaml \
        configs-hyphen-vars.yaml --save-files

With equivalent variables specified on the command line:

    sat bootprep run --dry-run \
        --vars "slingshot-host-software.version=2.0.0-75" \
        configs-hyphen-vars.yaml --save-files

## Summary and Scope

See commit message 

## Issues and Related PRs

Resolves CRAYSAT-1653

## Testing

See commit message

## Pull Request Checklist

- [X] Copyrights updated
- [X] License file intact
- [X] Target branch correct
- [X] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   man page.
 - Added ability to filter the images provided by a product using a prefix when
   specifying the base of an image in a `sat bootprep` input file.
+- Added support for hyphenated product names in `sat bootprep` variables.
 
 ## [3.20.0] - 2022-01-13
 
--- a/sat/cli/bootprep/vars.py
+++ b/sat/cli/bootprep/vars.py
@@ -1,7 +1,7 @@
 #
 # MIT License
 #
-# (C) Copyright 2022 Hewlett Packard Enterprise Development LP
+# (C) Copyright 2022-2023 Hewlett Packard Enterprise Development LP
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
 # copy of this software and associated documentation files (the "Software"),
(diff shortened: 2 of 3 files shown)


haasken-hpe (approved):
Looks good. Two minor comments:

* We should document this in docs-sat. I'm in favor of that being tracked under a separate Jira.
* Unit tests could be added pretty easily for this behavior I think.

ekamin-hpe:
> * We should document this in docs-sat. I'm in favor of that being tracked under a separate Jira.

https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1662

jack-stanek-hpe (approved):

ekamin-hpe:
> * Unit tests could be added pretty easily for this behavior I think.

Added a unit test with 94500fe5. This was a good idea because I discovered that the substitution was not happening when the `enumerate_vars_and_sources` method was being called (because it accesses each property individually rather than through `self.vars`). As a result, you would still see the hyphenated names in the output of `sat bootprep list-vars` which could be misleading.

haasken-hpe (approved):

PR #103, opened 2023-01-31
CRAYSAT-1604: Add configurable retries

ekamin-hpe:
## Summary and Scope

This commit adds two new options, `--api-retries` and `--api-backoff` that control the number of retries and interval between retries. These options can be specified via the config file or on the command line.

To implement these options, the urllib3.util.Retry class is used, which is passed to the requests.Session using an HTTPAdapter class. Retries will only be done for status codes in the 500 range or for network errors, errors like "400 Bad Request" or "403 Forbidden" will not result in any retries.

To ensure that retries are logged, handle log messages from the `urllib3` module.

This commit also documents the new command-line and config file options in the main sat(8) man page.

### Test Description:
* I ran `sat status` locally with an authenticated session to ensure normal usage was working.
* I temporarily moved my authentication token aside, resulting in 401 Unauthorized errors from the system when running `sat status`. I verified that it was not possible to add retries in this case.
* I modified my configuration file to specify an incorrect URL for the system's API gateway, resulting in network-level errors when running `sat status`. I verified that I was able to change both the number of retries and backoff factor both by using command-line options as well as configuration file options.
* I ran `sat bootprep` to create images, configurations and session templates on an internal system.

## Pull Request Checklist

- [X] Copyrights updated
- [X] License file intact
- [X] Target branch correct
- [X] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.35.4] - 2025-04-17
+
+### Added
+- Added configurable `api_gateway.retries` and `api_gateway.backoff` config
+  file parameters, and added command-line equivalents `--api-retries` and
+  `--api-backoff`.
+
 ## [3.35.3] - 2025-04-11
 
 ### Added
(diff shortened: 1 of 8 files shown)


ekamin-hpe:
Note to @jack-stanek-hpe and @haasken-hpe - I think the basics are working but I would like to test this in a bit more detail before merging. It seems as though vshasta may be a viable place to test this more thoroughly.

ekamin-hpe:
I'm finding a few issues in testing that I'm still trying to understand. Going to move this back to "draft" mode while I work it out.

vinanti-p (approved):

PR #110, opened 2023-02-23
CRAYSAT-1310: Capture and filter InsecureRequestWarning

ekamin-hpe:
## Summary and Scope

Add a new warnings.py module with a function to configure warnings.
* Uses 'once' filters and matches API gateway and S3 endpoint parameters
  so that we can get one warning per host.
* Add s3.cert_verify parameter but unfortunately it needs to remain
  False for now.
* Remove TODO references except in config.py where we should eventually
  change the default value of s3.cert_verify.
* Add missing documentation on s3 config file section to the man page.

## Testing

Test Description:
* I ran `sat status` with certificate verification enabled and disabled.
* I ran `sat showrev` with rgw certificate verification enabled and
  disabled.
* I visually inspected the resulting warnings.

## Pull Request Checklist

- [X] Copyrights updated
- [X] License file intact
- [X] Target branch correct
- [X] CHANGELOG.md updated

### DIFF
--- a/sat/cli/bootsys/state_recorder.py
+++ b/sat/cli/bootsys/state_recorder.py
@@ -1,7 +1,7 @@
 #
 # MIT License
 #
-# (C) Copyright 2020 Hewlett Packard Enterprise Development LP
+# (C) Copyright 2020,2023 Hewlett Packard Enterprise Development LP
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
 # copy of this software and associated documentation files (the "Software"),
@@ -29,7 +29,6 @@ import json
 import logging
 import os
 import sys
-from urllib3.exceptions import InsecureRequestWarning
 import warnings
 from datetime import datetime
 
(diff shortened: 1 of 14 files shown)


jack-stanek-hpe (approved):

sat/cli/bootsys/state_recorder.py
```diff
@@ -205,10 +204,7 @@ def dump_state(self):
             raise StateError(f'Failed to write state to file {local_new_file_name}: {err}') from err
         try:
             LOGGER.debug('Uploading %s to S3', new_file_name)
-            # TODO(SAT-926): Start verifying HTTPS requests
-            with warnings.catch_warnings():
-                warnings.filterwarnings('ignore', category=InsecureRequestWarning)
```
haasken-hpe, line 210: There are some similar `filterwarnings` calls in `sat.apiclient.ims.IMSClient` in the methods for renaming images. I imagine we could similarly get rid of the filtering there too now that we only print one warning for this. I might make that change and test it out with `sat bootprep`.

haasken-hpe (changes requested):
It seems like we should make this change in `sat.apiclient.ims` too, so I'm going to make that change and test it out before merging.

sat/cli/bootsys/state_recorder.py
haasken-hpe: Here is the gist with the test results from running `sat bootprep` against an input file that creates an image by configuring an existing image. This exercises the image rename code that uses the S3 API directly. As expected, there is just one warning despite multiple calls to S3 APIs. 

https://gist.github.com/haasken-hpe/5f7b003f1866b75438d3fd60974d43d3

haasken-hpe (approved):
Looks good to me. @jack-stanek-hpe, could you please give this another review for the additional changes I've made here?

haasken-hpe:
In testing, I noticed that if you enable cert verification on the IMS requests, you get a traceback in the `sat bootprep` case, so I just pushed a commit to catch that error appropriately. I will test again to make sure that fixes the traceback.

haasken-hpe:
I updated the gist with the results of testing. I had to switch systems because I was having trouble downloading the image to the system I was working on yesterday.  Here's a link to that gist again: https://gist.github.com/haasken-hpe/5f7b003f1866b75438d3fd60974d43d3

PR #114, opened 2023-03-22
CRAYSAT-1697: Add `wildcard` filtering option in bootprep input file

jack-stanek-hpe:
## Summary and Scope

This change adds a `wildcard` option to product-supplied images in bootprep input files. The `wildcard` option can be specified under a `base.product.filter` clause to select the base image from a product. Wildcard pattern matching is performed using standard glob syntax. See documentation of the Python `fnmatch` module for more information.

## Issues and Related PRs

* Resolves [CRAYSAT-1697](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1697)

### Tested on:

  * `rocket`
  
### Test description:

Run unit tests. Run on a system against a simple bootprep input file with a single COS image selected using a wildcard.

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated

### DIFF
--- a/sat/cli/bootprep/input/image.py
+++ b/sat/cli/bootprep/input/image.py
@@ -25,6 +25,7 @@
 Defines class for images defined in the input file.
 """
 from abc import ABC, abstractmethod
+import fnmatch
 import logging
 import warnings
 
@@ -640,9 +641,16 @@ class ProductInputImage(BaseInputImage):
         return self.image_data['base']['product'].get('version')
 
     @property
+    @jinja_rendered
     def filter_prefix(self):
         """str or None: the filter prefix to use for filtering images or recipes from the product"""
         return get_val_by_path(self.image_data, 'base.product.filter.prefix')
+
+    @property
+    @jinja_rendered
+    def filter_wildcard(self):
+        """str or None: the filter pattern to use for filtering images or recipes from the product"""
+        return get_val_by_path(self.image_data, 'base.product.filter.wildcard')
 
     def filter_func(self, name):
         """Filter the given name.
@@ -653,7 +661,15 @@ class ProductInputImage(BaseInputImage):
         Returns:
             True if the filter matches the given name or there is no filter, False otherwise
         """
-        return self.filter_prefix is None or name.startswith(self.filter_prefix)
+        # Schema validation guarantees that there cannot be both a prefix and a wildcard,
+        # so it is possible to just check each individually and not worry about the case
+        # where both are present.
+        if self.filter_prefix:
+            return name.startswith(self.filter_prefix)
+        if self.filter_wildcard is not None:
+            return fnmatch.fnmatch(name, self.filter_wildcard)
+
+        return True
 
     @property
     def unfiltered_base_description(self):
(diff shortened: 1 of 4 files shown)


sat/data/schema/bootprep_schema.yaml
```diff
@@ -275,7 +275,7 @@ properties:
                         Filter the image or recipe to use from the given version of the given product
                         in case this product provides more than one image or recipe and one must be
                         selected.
-                      required: ['prefix']
+                      maxProperties: 1
```
haasken-hpe, line 278: Nice, TIL. Should we also specify `minProperties: 1` to force it to be exactly one property so they can't do something like:

```
filter: {}
```

It looks like our code would handle that case just the same as if they hadn't specified `filter` at all, but maybe preventing it in the schema would be smart to reduce the likelihood of errors?

sat/cli/bootprep/input/image.py
haasken-hpe, line 1: Just a reminder about the second UAC in the Jira which is to make the `filter_prefix` Jinja2 rendered so that something like `{{cos.version}}` could be specified in that field. I think all that's required to accomplish that is to add the `@jinja_rendered` decorator on the `filter_prefix` method. I think we could also add that to the `filter_wildcard` for consistency and in case it's useful for some reason. I don't think the Jinja syntax should conflict with the wildcard syntax, but I could be wrong.

Also add a note to the `filter.prefix` in the schema to note that it is a Jinja2-rendered field like I did for the others.

sat/data/schema/bootprep_schema.yaml
haasken-hpe, line 1: Please bump the patch component of the version on line 44 for this fully backwards-compatible change.

haasken-hpe (changes requested):
The `filter.wildcard` implementation looks good, but the second UAC about adding rendering to the `filter.prefix` is missing. My other minor comments should be easy to address at the same time.

sat/data/schema/bootprep_schema.yaml
jack-stanek-hpe: Seems reasonable, I've added this to the schema.

haasken-hpe (approved):

sat/cli/bootprep/input/image.py
```diff
@@ -653,7 +659,15 @@ def filter_func(self, name):
         Returns:
             True if the filter matches the given name or there is no filter, False otherwise
         """
-        return self.filter_prefix is None or name.startswith(self.filter_prefix)
+        # Schema validation guarantees that there cannot be both a prefix and a wildcard,
+        # so it is possible to just check each individually and not worry about the case
+        # where both are present.
+        if self.filter_prefix:
+            return name.startswith(self.filter_prefix)
+        if self.filter_wildcard:
```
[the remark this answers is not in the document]
haasken-hpe: Looks like it will match nothing now, which I'm fine with. If they did specify an empty string for `filter.wildcard`, it's gotta be a mistake.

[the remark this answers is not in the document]
haasken-hpe: Ahh yes, that's a good idea.

haasken-hpe (approved):

PR #119, opened 2023-04-07
CRAYSAT-1705: Fix logic for checking when staged BOS session is complete

haasken-hpe:
## Summary and Scope

Normally, a BOS v2 staged session will reach the state "running" when it is considered finished staging. However, if that BOS v2 staged session does not apply to any components (for example, when a limit parameter is used that does not overlap with any of the nodes in the session template), the session will go straight to "complete" instead.

Rename the `target_state` argument to the BOSV2SessionWaiter to `target_states` and change the type to a list of acceptable target states. This allows us to call it with the single state "complete" when waiting on a non-staged session and to call it with the states "complete" and "running" when waiting on a staged session.

## Issues and Related PR

* Resolves [CRAYSAT-1705](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1705)

### Tested on:

  * locally on macbook
  * TBD

### Test description:

Added new unit tests. Will test on a real system as follows:

* Run `sat bootsys boot --stage bos-operations --staged-session` using a session template that applies to application nodes, but pass `--bos-limit Compute`. The session should finish and `sat bootsys` should return.
* Run the same test as above but pass two session templates: one for compute nodes and one for UANs. Both should finish and `sat bootsys` should return. Only computes should be staged.
* Run the same test as above but without `--staged-session`. The computes should actually be rebooted.

## Risks and Mitigations

This is a pretty small change that just adds an additional acceptable target state for staged BOS v2 sessions.

Unit tests cover the changed code.

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated

### DIFF
--- a/sat/cli/bootsys/bos.py
+++ b/sat/cli/bootsys/bos.py
@@ -212,19 +212,28 @@ def boa_job_successful(boa_job_id):
 class BOSV2SessionWaiter(Waiter):
     """Waiter for monitoring the status of a BOS v2 session."""
 
-    def __init__(self, bos_session_thread, target_state='complete', *args, **kwargs):
+    def __init__(self, bos_session_thread, target_states=None, *args, **kwargs):
+        """
+        Create a new BOSV2SessionWaiter.
+
+        Args:
+            bos_session_thread (BOSSessionThread): the BOS session thread
+            target_states (list, optional): the list of acceptable target state
+                strings. Defaults to ['complete']
+        """
         super().__init__(*args, **kwargs)
         self.bos_client = BOSClientCommon.get_bos_client(SATSession(),
                                                          version='v2')
         self.bos_session_thread = bos_session_thread
-        self.target_state = target_state
+        self.target_states = ['complete'] if target_states is None else target_states
+        self.target_states_desc = INFLECTOR.join(self.target_states, conj='or')
 
         self.pct_successful = 0.0
         self.pct_failed = 0.0
         self.error_summary = {}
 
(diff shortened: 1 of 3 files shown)


sat/cli/bootsys/bos.py
```diff
@@ -253,12 +263,11 @@ def has_completed(self):
             if session_status['error_summary']:
                 self.error_summary = session_status['error_summary']
 
-            if session_status['status'] == self.target_state:
+            if session_status['status'] in self.target_states:
```
haasken-hpe, line 266: There was a conflict here due to a change in integration that we haven't backported to release/3.21. Can you double-check that I handled it correctly, @jack-stanek-hpe?

jsollom-hpe (approved):

haasken-hpe:
Testing is looking good so far. I am just waiting for the actual reboot of computes to complete:

https://gist.github.com/haasken-hpe/0a420567bc169800fee4210380baf1ad

jack-stanek-hpe (approved):

haasken-hpe:
> Testing is looking good so far. I am just waiting for the actual reboot of computes to complete:
> 
> https://gist.github.com/haasken-hpe/0a420567bc169800fee4210380baf1ad

I realized that 3 of the computes were not booted before, so this session will probably not actually reach a "complete" status in BOS. I think it adequately tested up to this point though.

PR #130, opened 2023-05-15
CRAYSAT-654: Recreate cronjobs which are not being scheduled

jack-stanek-hpe:
## Summary and Scope

This change implements a feature in the cabinet power stage of power on which recreates Kubernetes cronjobs if they are not being scheduled according to their cron schedule. This is determined by checking if the cron job's last scheduled time was prior to the computed prior scheduled time based on its schedule. All cronjobs which are stuck not executing will be deleted and recreated with the same specification.

## Issues and Related PRs

* Resolves [CRAYSAT-654](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-654)

### Tested on:

### Test description:

Run unit tests. Create cron job which sleeps for longer than its cron schedule with a Forbid concurrency policy (thus causing it to frequently miss its scheduled time.) Run a simple test harness which contains the stuck cronjob check loop from the cabinet power stage and ensure it deletes and recreates the cron jobs.

See https://gist.github.com/jack-stanek-hpe/778b409affc120ff75b4fa8c0e045160 for simple manual testing.

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated

### DIFF
--- a/sat/cli/bootsys/cabinet_power.py
+++ b/sat/cli/bootsys/cabinet_power.py
@@ -26,13 +26,18 @@ Powers on and off liquid-cooled compute cabinets.
 """
 import logging
 
+from csm_api_client.k8s import load_kube_api
+from kubernetes.client import BatchV1Api
+from kubernetes.config import ConfigException
+
 from sat.apiclient import APIError, CAPMCClient, HSMClient
 from sat.cli.bootsys.power import CAPMCPowerWaiter
 from sat.config import get_config_value
-from sat.hms_discovery import HMSDiscoveryCronJob, HMSDiscoveryError, HMSDiscoveryScheduledWaiter
+from sat.cronjob import recreate_namespaced_stuck_cronjobs
+from sat.hms_discovery import (HMSDiscoveryCronJob, HMSDiscoveryError,
+                               HMSDiscoveryScheduledWaiter)
 from sat.session import SATSession
 from sat.util import prompt_continue
-
 
 LOGGER = logging.getLogger(__name__)
 
@@ -200,6 +205,13 @@ def do_cabinets_power_on(args):
         LOGGER.error(f'Failed to start waiting on HMS discovery job being '
                      f'rescheduled after resume: {err}')
         raise SystemExit(1)
+
+    LOGGER.info('Recreating cronjobs which have failed to have been scheduled on time.')
+    try:
+        batch_api = load_kube_api(api_cls=BatchV1Api)
+        recreate_namespaced_stuck_cronjobs(batch_api, 'services')
+    except ConfigException as err:
+        LOGGER.warning('Could not load Kubernetes configuration: %s', err)
 
     hms_discovery_scheduled = hms_discovery_waiter.wait_for_completion()
     if not hms_discovery_scheduled:
--- a/sat/cli/bootsys/platform.py
+++ b/sat/cli/bootsys/platform.py
@@ -1,7 +1,7 @@
 #
 # MIT License
 #
-# (C) Copyright 2021 Hewlett Packard Enterprise Development LP
+# (C) Copyright 2021, 2023 Hewlett Packard Enterprise Development LP
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
 # copy of this software and associated documentation files (the "Software"),
@@ -25,14 +25,17 @@
 Start and stop platform services to boot and shut down a Shasta system.
 """
 
-from collections import namedtuple
 import logging
 import socket
-from paramiko import SSHException
+from collections import namedtuple
 from threading import Thread
 
+from csm_api_client.k8s import load_kube_api
+from kubernetes.client import BatchV1Api
+from kubernetes.config import ConfigException
+from paramiko import SSHException
+
 from sat.cached_property import cached_property
-from sat.config import get_config_value
 from sat.cli.bootsys.ceph import (
     check_ceph_health,
     toggle_ceph_freeze_flags,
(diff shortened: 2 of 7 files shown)


sat/cli/bootsys/cabinet_power.py
```diff
@@ -202,6 +293,13 @@ def do_cabinets_power_on(args):
                      f'within expected window after being resumed.')
         raise SystemExit(1)
 
+    LOGGER.info('Recreating cronjobs which have failed to have been scheduled on time.')
```
haasken-hpe, line 296: I think we want to put this after the hms-discovery cronjob is resumed and before we wait on it to be scheduled. Otherwise, that waiting will time out when the hms-discovery cronjob needs to be re-created.

And thanks to the way you've structured this code, it should also enable us to pull the new cronjob functions into their own `cronjob` module and then perhaps call `recreate_namespaced_stuck_cronjobs` here in the cabinet-power stage and also call it at the end of the platform-services stage. That would be a little more consistent with the ordered procedure in the docs today, which has the admin check the cronjobs during [Power On and Start the Kubernetes Management Cluster](https://github.com/Cray-HPE/docs-csm/blob/main/operations/power_management/Power_On_and_Start_the_Management_Kubernetes_Cluster.md#check-cronjobs) procedure and then again specifically for hms-discovery in step 4 of the [Power On Compute Cabinets](https://github.com/Cray-HPE/docs-csm/blob/main/operations/power_management/Power_On_Compute_Cabinets.md#power-on-cray-ex-liquid-cooled-cabinet-circuit-breakers) procedure.

tests/cli/bootsys/test_cabinet_power.py
```diff
@@ -69,3 +78,95 @@ def test_do_cabinets_power_off_without_prompting(self):
 
         self.mock_liquid_cooled.assert_called_once_with(self.args)
         self.mock_air_cooled.assert_called_once_with(self.args)
+
+
+class TestRecreateCronjobs(unittest.TestCase):
+    def setUp(self):
+        now = datetime.datetime.now(tz=tzutc())
+        self.now = datetime.datetime(*now.timetuple()[:4], tzinfo=tzutc())
```
haasken-hpe, line 86: A comment here that this is getting the current "top of the hour" time would be helpful

haasken-hpe (changes requested):
Requesting changes for my comment about the order of re-creating cronjobs relative to waiting on hms-discovery. It should be an easy change.

sat/cli/bootsys/cabinet_power.py
```diff
@@ -24,19 +24,110 @@
...
+            with the Kubernetes API
+    """
+    cronjob = copy.deepcopy(cronjob)
+    name = cronjob.metadata.name
+    namespace = cronjob.metadata.namespace
+    cronjob.metadata.resource_version = None
+    try:
+        batch_api.delete_namespaced_cron_job(name, namespace)
+    except ApiException as err:
+        if err.status != 404:
```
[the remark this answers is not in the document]
jack-stanek-hpe: This is sort of an artifact of my testing, as I was initially passing in the name and namespace as arguments to `recreate_cronjob()` instead of the `V1CronJob` object. I think it still can serve a purpose, since IMO if we "recreate" a cronjob that doesn't exist, the logical end result would be that the cron job is created after all.

I also wanted to be a bit defensive here since the recreation operation isn't atomic; if for any reason this function fails between deleting and recreating the cronjob, then there could be a situation where _no_ cron job exists, and I wanted to smooth over an easy-to-handle error case.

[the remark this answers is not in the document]
haasken-hpe: That makes sense, thanks.

haasken-hpe (approved):
The order change in cabinet-power looks good, but I can review again when the cronjob re-creation is added to the end of the platform-services stage.

sat/cli/bootsys/cabinet_power.py
```diff
@@ -193,6 +198,13 @@ def do_cabinets_power_on(args):
         LOGGER.error(f'Failed to resume discovery: {err}')
         raise SystemExit(1)
 
+    LOGGER.info('Recreating cronjobs which have failed to have been scheduled on time.')
```
haasken-hpe, line 201: I realized that this is now probably always going to re-create the hms-discovery cronjob because it will not have been scheduled yet immediately after it's been resumed. I discussed with Jack, and he's going to see about putting it in the `on_retry_action`.

sat/cli/bootsys/platform.py
```diff
@@ -464,6 +469,16 @@ def do_kubelet_start(ncn_groups):
                                target_state='active', target_enabled='enabled')
 
 
+def do_recreate_cronjobs(_):
+    """Recreate cronjobs that are not being scheduled on time."""
+    LOGGER.info('Recreating cronjobs which have failed to have been scheduled on time.')
```
haasken-hpe, line 474: This log message is redundant with the info-level log message emitted on line 616 in `do_platform_action`.

sat/cronjob.py
```diff
@@ -0,0 +1,119 @@
...
+        if cronjob_stuck(cronjob):
+            LOGGER.warning('Jobs for cronjob "%s" in namespace "%s" do not appear to be '
+                           'scheduled on time according to the cron job\'s schedule; '
+                           'recreating cron job.',
+                           cronjob.metadata.name, cronjob.metadata.namespace)
+            try:
+                recreate_cronjob(batch_api, cronjob)
+            except ApiException as err:
+                LOGGER.warning('An error occurred while re-creating cronjob "%s" in namespace "%s": %s',
+                               cronjob.metadata.name, cronjob.metadata.namespace, err)
```
haasken-hpe, line 119: PEP8 wants a newline at the end of this file

sat/cli/bootsys/cabinet_power.py
```diff
@@ -201,6 +206,13 @@ def do_cabinets_power_on(args):
                      f'rescheduled after resume: {err}')
         raise SystemExit(1)
 
+    LOGGER.info('Recreating cronjobs which have failed to have been scheduled on time.')
+    try:
+        batch_api = load_kube_api(api_cls=BatchV1Api)
+        recreate_namespaced_stuck_cronjobs(batch_api, 'services')
+    except ConfigException as err:
+        LOGGER.warning('Could not load Kubernetes configuration: %s', err)
+
```
haasken-hpe, line 215: Now that you are re-creating a possibly stuck hms-discovery cronjob in the `HMSDiscoveryScheduledWaiter` in the `on_retry_action`, you wouldn't want to unconditionally re-create the cronjob here, would you?

PR #131, opened 2023-05-16
CRAYSAT-1623: Improve hms-discovery scheduled check

jack-stanek-hpe:
## Summary and Scope

This change modifies the logic in the HMSDiscoveryScheduledWaiter to
check which Kubernetes jobs exist for the hms-discovery cronjob, and
checks if their creation timestamp is after the time that waiting began.
If it was, then waiting finishes; if no job is created within the
timeout duration, then the waiter will time out as normal.

A fix is also included to update the location of kubectl version in the metal-provision repo.

## Issues and Related PRs

* Resolves [CRAYSAT-1623](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1623)
* Also resolves [CRAYSAT-1631](https://jira-pro.its.hpecorp.net:8443/browse/CRAYSAT-1631)

### Tested on:

### Test description:

Create stub version of hms-discovery cron job in minikube. Use waiter to
wait until the job is scheduled. Modify cron job such that it will never
schedule, and make sure that the waiter times out.

Run unit tests.

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated

### DIFF
--- a/sat/hms_discovery.py
+++ b/sat/hms_discovery.py
@@ -24,18 +24,18 @@
 """
 Functionality related to the hms-discovery cronjob in k8s.
 """
-from datetime import datetime
 import logging
+from datetime import datetime
 
 from croniter import croniter
 from csm_api_client.k8s import load_kube_api
 from dateutil.tz import tzutc
-from kubernetes.config import ConfigException
-from kubernetes.client import BatchV1beta1Api
+from kubernetes.client import BatchV1Api
 from kubernetes.client.rest import ApiException
+from kubernetes.config import ConfigException
 
 from sat.cached_property import cached_property
-from sat.waiting import Waiter
+from sat.waiting import Waiter, WaitingFailure
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


sat/hms_discovery.py
```diff
@@ -209,19 +230,17 @@ def has_completed(self):
...
-            return True
-        else:
-            return False
+            jobs = self.hd_cron_job.get_jobs()
+            return any(
+                job.metadata.creation_timestamp > self.start_time
+                for job in jobs
+            )
+        except ApiException as err:
+            raise WaitingFailure(
```
haasken-hpe, line 239: This is a slight change in behavior from before, and I just want to make sure we discuss the rationale.

Before, if there was some failure when querying the K8s API for the cronjob's last schedule time, it raised `HMSDiscoveryError` from the K8s `ApiException`, and then we caught it, logged a warning (on line 214 on the left), and then returned False, so it would keep trying until the timeout.

With this change, if there is a failure listing cronjobs through the K8s API, we will get a K8s `ApiException` from `get_jobs`, which is caught, and then we raise a `WaitingFailure` which causes the waiting to stop.

It seems fairly unlikely that we will encounter such an error, but if we do, it seems like it might be beneficial to tolerate it if it's a transient problem. Thoughts?

haasken-hpe (approved):
My comments are pretty minor.

sat/hms_discovery.py
jack-stanek-hpe: My reasoning was that if the jobs aren't able to be retrieved, then it may be less likely to be a transient error. However, reviewing the previous implementation, the `HMSDiscoveryError` was being raised if the `hms-discovery` cronjob couldn't be read from the K8s API, which is probably the same level of severity. I can revert to continuing to wait, I don't feel strongly about stopping waiting on error.

PR #137, opened 2023-06-07
CRAYSAT-1722: Speed up known_hosts loading

jack-stanek-hpe:
## Summary and Scope

This change filters the known_hosts file prior to loading it into a paramiko SSHClient object in order to improve performance. Paramiko's implementation of known_hosts parsing has quadratic time complexity, which causes very poor performance in general when a large known_hosts file is used. In our particular case, it was causing additional slowdown due to the GIL since we were using the `threading` module and parsing the known hosts file several times on different threads, causing GIL thrashing.

To address this, this change pre-filters the known_hosts file to just the hosts needed in a given stage, and also changes the container stop threads to use `multiprocessing` instead, so that multiple SSH connections can occur simultaneously without compromising (as much) performance.

## Issues and Related PRs

* Resolves [CRAYSAT-1710](https://jira-pro.it.hpe.com:8443/browse/CRAYSAT-1710)

### Tested on:

### Test description:
Run unit tests. Use benchmark test harness to check runtimes of filtered host keys to ensure performance is at least 1 order of magnitude improved.

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated

### DIFF
--- a/sat/cli/bootsys/etcd.py
+++ b/sat/cli/bootsys/etcd.py
@@ -1,7 +1,7 @@
 #
 # MIT License
 #
-# (C) Copyright 2021 Hewlett Packard Enterprise Development LP
+# (C) Copyright 2021, 2023 Hewlett Packard Enterprise Development LP
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
 # copy of this software and associated documentation files (the "Software"),
@@ -52,11 +52,13 @@ class EtcdInactiveFailure(EtcdSnapshotFailure):
     pass
 
 
-def save_etcd_snapshot_on_host(hostname):
+def save_etcd_snapshot_on_host(hostname, host_keys=None):
     """Connect to the given host and save an etcd snapshot to a file.
 
     Args:
         hostname (str): the hostname to connect to
+        host_keys (paramiko.HostKeys or None): the hostkeys to use when
+            connecting over SSH
 
     Raises:
         EtcdInactiveFailure: if the etcd service is inactive, indicating that
@@ -64,7 +66,7 @@ def save_etcd_snapshot_on_host(hostname):
         EtcdSnapshotFailure: if there is a failure to create the directory for
             the snapshot or a failure to create the snapshot
     """
-    ssh_client = get_ssh_client()
+    ssh_client = get_ssh_client(host_keys=host_keys)
 
     try:
         ssh_client.connect(hostname)
--- a/sat/cli/bootsys/platform.py
+++ b/sat/cli/bootsys/platform.py
@@ -28,7 +28,7 @@ Start and stop platform services to boot and shut down a Shasta system.
 import logging
 import socket
 from collections import namedtuple
-from threading import Thread
+from multiprocessing import Process
 
 from csm_api_client.k8s import load_kube_api
 from kubernetes.client import BatchV1Api
@@ -43,6 +43,7 @@ from sat.cli.bootsys.ceph import (
     CephHealthWaiter
 )
 from sat.cli.bootsys.etcd import save_etcd_snapshot_on_host, EtcdInactiveFailure, EtcdSnapshotFailure
+from sat.cli.bootsys.hostkeys import FilteredHostKeys
 from sat.cli.bootsys.util import get_and_verify_ncn_groups, get_ssh_client, FatalBootsysError
 from sat.config import get_config_value
 from sat.cronjob import recreate_namespaced_stuck_cronjobs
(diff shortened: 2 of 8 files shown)


sat/cli/bootsys/platform.py
```diff
@@ -377,6 +386,54 @@ def run(self):
...
+            tmp_known_hosts.flush()
+            super().load(tmp_known_hosts.name)
+
+    def save(self, filename):
+        """(Don't) save the known_hosts file.
+
+        This is a no-op since we don't want to accidentally destructively
+        modify the user's known hosts file.
+        """
+        return NotImplemented
```
haasken-hpe, line 434: Did you perhaps mean to `raise NotImplementedError`? Or perhaps this should just be an error log message?

tests/cli/bootsys/test_platform.py
```diff
@@ -33,6 +35,8 @@
 import unittest
 from unittest import mock
 
+from paramiko.hostkeys import HostKeyEntry
```
haasken-hpe, line 38: Unused import. `WarningPolicy` (line 32) is also unused.

sat/cli/bootsys/platform.py
haasken-hpe, line 1: We also need to add the known_hosts filtering to the `do_etcd_snapshot` function since that also uses Paramiko to run remote commands via SSH.

haasken-hpe (changes requested):
The most important items that should be addressed before merge would be:

* Closing SSHClient before we're done using it in `ContainerStopThread`
* Not filtering known hosts in the step which takes an etcd snapshot, which means that step will still be painfully slow.

sat/cli/bootsys/platform.py
```diff
@@ -220,16 +224,19 @@ def has_completed(self):
         return current_state == self.target_state
 
 
-class ContainerStopThread(Thread):
+class ContainerStopThread(Process):
```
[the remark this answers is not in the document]
jack-stanek-hpe: SSH protocol handling also happens within the Python threads, so we are impeded by the GIL there as well, so we should expect to see some amount of speedup.

sat/cli/bootsys/platform.py
jack-stanek-hpe: I actually thought before that `NotImplemented` was just a singleton without special properties, akin to `None` or `Ellipsis`, but apparently `return NotImplemented` does have special meaning. According to [the docs](https://docs.python.org/3/library/constants.html#NotImplemented), if it's returned by a special binary operator, e.g. `__eq__()` or `__le__()`, then CPython will try the other type's corresponding operation (e.g. `__eq__()` reflexively or `__gt__()`), so for non magic methods it can cause weird error messages. I'll just change it to return `None`, I don't want to raise exceptions in a method other library code can use.

sat/cli/bootsys/platform.py
```diff
@@ -377,6 +386,54 @@ def run(self):
...
 
+class FilteredHostKeys(HostKeys):
+    """
+    Host keys loader which filters all hosts except the specified ones.
+
+    This is needed to speed up known_hosts file parsing as paramiko's implementation
+    (up to and including 3.2.0) has quadratic time complexity. To get around this,
+    we can hackily cut down a large known hosts file to just a few hosts in linear
+    time before parsing, which should help tremendously in the case of a very large
+    known_hosts file.
```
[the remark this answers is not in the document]
jack-stanek-hpe: That's true, I'll add that to the comment.

sat/cli/bootsys/platform.py
haasken-hpe: Yeah, that seems reasonable. I'm not expecting we will actually save the hosts file anywhere, but I haven't dug through paramiko to see how this class is used. I wonder if we could also log a warning or something.

sat/cli/bootsys/etcd.py
```diff
@@ -28,6 +28,7 @@
 import os
 import socket
 
+from sat.cli.bootsys.hostkeys import FilteredHostKeys
```
haasken-hpe, line 31: Unused import

sat/cli/bootsys/platform.py
```diff
@@ -220,16 +223,19 @@ def has_completed(self):
         return current_state == self.target_state
 
 
-class ContainerStopThread(Thread):
+class ContainerStopProcess(Process):
     """A thread that will stop containers on hosts."""
```
haasken-hpe, line 227: Missed a "thread" here.

sat/cli/bootsys/platform.py
haasken-hpe, line 1: Pretty minor, but they're still called threads instead of processes in `do_stop_containers`.

haasken-hpe (approved):
Minor comments.

haasken-hpe (approved):

PR #161, opened 2023-08-08
CRAYSAT-1676: showrev bug fix when run without option

shivaprasad-metimath:
## Summary and Scope

error exception is added to handle and continue instead of aborting the cmd run

## Issues and Related PRs

* Resolves [CRAYSAT-1676](https://jira-pro.it.hpe.com:8443/browse/)

### Tested on:

  * will be tested on csm system

### Test description:

TODO

## Risks and Mitigations

Low risk.

## Pull Request Checklist

- [ ] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.25.4] - 2023-09-01
+
+### Fixed
+- Update `sat showrev` to log a warning and proceed to display other system
+  revision information when S3 credentials are not configured.
+- Update `sat setrev` and `sat bootsys` to handle missing S3 credentials more
+  gracefully.
+
 ## [3.25.3] - 2023-08-31
 
 ### Fixed
--- a/sat/cli/bootsys/state_recorder.py
+++ b/sat/cli/bootsys/state_recorder.py
@@ -45,7 +45,7 @@ from sat.cli.bootsys.defaults import (
 )
 from sat.cli.bootsys.util import k8s_pods_to_status_dict
 from sat.config import get_config_value
-from sat.util import BeginEndLogger, get_s3_resource
+from sat.util import BeginEndLogger, get_s3_resource, S3ResourceCreationError
 
 
 LOGGER = logging.getLogger(__name__)
@@ -321,20 +321,24 @@ def do_state_capture(args):
     Raises:
         SystemExit: if there is a failure to capture state.
     """
-    state_recorders = [PodStateRecorder()]
-
-    failed = []
-    LOGGER.info('Capturing system state.')
-    with BeginEndLogger('system state capture'):
-        for sr in state_recorders:
-            try:
-                LOGGER.info(f'Capturing {sr.description}')
-                sr.dump_state()
-            except StateError as err:
-                LOGGER.error(f'Failed to capture {sr.description}: {err}')
-                failed.append(sr)
-
-    if failed:
+    try:
+        state_recorders = [PodStateRecorder()]
+    except S3ResourceCreationError as err:
+        LOGGER.error(f'Failed to initialize PodStateRecorder: {err}')
         sys.exit(1)
     else:
-        LOGGER.info('Finished capturing system state.')
+        failed = []
+        LOGGER.info('Capturing system state.')
+        with BeginEndLogger('system state capture'):
+            for sr in state_recorders:
+                try:
+                    LOGGER.info(f'Capturing {sr.description}')
+                    sr.dump_state()
+                except StateError as err:
+                    LOGGER.error(f'Failed to capture {sr.description}: {err}')
+                    failed.append(sr)
+
+        if failed:
+            sys.exit(1)
+        else:
+            LOGGER.info('Finished capturing system state.')
--- a/sat/cli/setrev/main.py
+++ b/sat/cli/setrev/main.py
@@ -35,8 +35,7 @@ import yaml
 
 from sat.cli.setrev.site_fields import SITE_FIELDS
 from sat.config import get_config_value
-from sat.util import get_s3_resource, yaml_dump
-
+from sat.util import get_s3_resource, yaml_dump, S3ResourceCreationError
 
 LOGGER = logging.getLogger(__name__)
 
@@ -54,18 +53,21 @@ def get_site_data(sitefile):
     Raises:
         PermissionError: If the sitefile exists locally but could not be read.
     """
-    s3 = get_s3_resource()
-    s3_bucket = get_config_value('s3.bucket')
-
     try:
-        LOGGER.debug('Downloading %s from S3 (bucket: %s)', sitefile, s3_bucket)
-        s3.Object(s3_bucket, sitefile).download_file(sitefile)
-    except (BotoCoreError, ClientError, Boto3Error) as err:
-        # It is not an error if this file doesn't already exist
-        # TODO: it would be nice to differentiate between successfully connecting to S3
-        # and the file not existing versus failing to connect to S3, where the latter
-        # deserves at least a warning-level message.
-        LOGGER.debug('Unable to download site info file %s from S3: %s', sitefile, err)
+        s3 = get_s3_resource()
+    except S3ResourceCreationError as err:
+        LOGGER.warning('Error creating S3 resource: %s', err)
+    else:
+        s3_bucket = get_config_value('s3.bucket')
+        try:
+            LOGGER.debug('Downloading %s from S3 (bucket: %s)', sitefile, s3_bucket)
+            s3.Object(s3_bucket, sitefile).download_file(sitefile)
+        except (BotoCoreError, ClientError, Boto3Error) as err:
+            # It is not an error if this file doesn't already exist
+            # TODO: it would be nice to differentiate between successfully connecting to S3
+            # and the file not existing versus failing to connect to S3, where the latter
+            # deserves at least a warning-level message.
+            LOGGER.debug('Unable to download site info file %s from S3: %s', sitefile, err)
 
     try:
         with open(sitefile, 'r') as f:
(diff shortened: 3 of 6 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.25.3] - 2023-08-08
+
+###Bug fix
```
haasken-hpe, line 30:
```suggestion
### Fixed
```

sat/cli/setrev/main.py
```diff
@@ -54,18 +53,23 @@ def get_site_data(sitefile):
...
-    except (BotoCoreError, ClientError, Boto3Error) as err:
-        # It is not an error if this file doesn't already exist
-        # TODO: it would be nice to differentiate between successfully connecting to S3
-        # and the file not existing versus failing to connect to S3, where the latter
-        # deserves at least a warning-level message.
-        LOGGER.debug('Unable to download site info file %s from S3: %s', sitefile, err)
+        s3 = get_s3_resource()
+    except S3ResourceCreationError as err:
+        LOGGER.error('Error creating S3 resource: %s', err)
+        s3 = None  # Set s3 to None to skip S3 operations
```
haasken-hpe, line 60: I think I'd consider this a warning. Also, there will be redundant log messages because currently `get_s3_resource` logs an error, and then so does this function. I think it would be better to just log the warning here.

I've also elaborated on this in my comments on `get_s3_resource`.

sat/cli/setrev/main.py
```diff
@@ -54,18 +53,23 @@ def get_site_data(sitefile):
...
+    if s3:
+        try:
+            LOGGER.debug('Downloading %s from S3 (bucket: %s)', sitefile, s3_bucket)
+            s3.Object(s3_bucket, sitefile).download_file(sitefile)
+        except (BotoCoreError, ClientError, Boto3Error) as err:
+            # It is not an error if this file doesn't already exist
+            # TODO: it would be nice to differentiate between successfully connecting to S3
+            # and the file not existing versus failing to connect to S3, where the latter
+            # deserves at least a warning-level message.
+            LOGGER.debug('Unable to download site info file %s from S3: %s', sitefile, err)
```
haasken-hpe, line 72: You could just use an `else` on the try/except like this:

```suggestion
    else:
        s3_bucket = get_config_value('s3.bucket')
        try:
            LOGGER.debug('Downloading %s from S3 (bucket: %s)', sitefile, s3_bucket)
            s3.Object(s3_bucket, sitefile).download_file(sitefile)
        except (BotoCoreError, ClientError, Boto3Error) as err:
            # It is not an error if this file doesn't already exist
            # TODO: it would be nice to differentiate between successfully connecting to S3
            # and the file not existing versus failing to connect to S3, where the latter
            # deserves at least a warning-level message.
            LOGGER.debug('Unable to download site info file %s from S3: %s', sitefile, err)
```

sat/cli/setrev/main.py
```diff
@@ -82,6 +86,7 @@ def get_site_data(sitefile):
                        'It will be replaced if you continue.', sitefile)
         return {}
 
+
```
haasken-hpe, line 89: PEP 8: remove extra blank line

sat/util.py
```diff
@@ -661,12 +662,12 @@ def get_s3_resource():
             )
     except OSError as err:
         LOGGER.error(f'Unable to load configuration: {err}')
-        raise SystemExit(1)
+        raise S3ResourceCreationError(str(err))
```
haasken-hpe, line 665: I find that error and exception handling is a delicate art. Generally, I think if a function is going to raise an exception to the caller when a problem is encountered, it is not usually necessary to also log an error message. The caller may decide that this error condition is non-fatal and may want to log a warning and continue instead. In fact, I think that is the case for `sat showrev`.

If we get rid of the `LOGGER.error`, it would be helpful to add context to the error message passed to the constructor of `S3ResourceCreationError`. It's also possible to keep a reference to the original exception in the new exception. Putting all that together, I'd suggest something like this:

```suggestion
        raise S3ResourceCreationError(f'Unable to load S3 key: {err}') from err
```

Here's an example in my interpreter of the logged errors before my change:

```
>>> try:
...   s3 = get_s3_resource()
... except Exception as err:
...   print(err)
...
Unable to load configuration: [Errno 2] No such file or directory: '/Users/haasken/.config/sat/my_s3_access_key'
[Errno 2] No such file or directory: '/Users/haasken/.config/sat/my_s3_access_key'
```

And after my suggested change:

```
>>> from sat.util import get_s3_resource
>>> try:
...   s3 = get_s3_resource()
... except Exception as err:
...   print(err)
...
Unable to load S3 key: [Errno 2] No such file or directory: '/Users/haasken/.config/sat/my_s3_access_key'
```

sat/util.py
```diff
@@ -661,12 +662,12 @@ def get_s3_resource():
...
     except ValueError as err:
         # A ValueError is raised when an invalid value is given to boto3.resource,
         # e.g. an endpoint URL that is not a valid URL.
         LOGGER.error(f'Unable to load S3 API: {err}')
-        raise SystemExit(1)
+        raise S3ResourceCreationError(str(err))
```
haasken-hpe, line 670: Similar to the above comment:

```suggestion
        raise S3ResourceCreationError(f'Unable to load S3 API: {err}') from err
```

haasken-hpe (commented):
The overall approach looks good here. I have a few small comments to address, and I think you can continue making similar changes to the other callers of `get_s3_resource`.

sat/util.py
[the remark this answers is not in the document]
shivaprasad-metimath: addressed at line no 648

docker_scripts/config-docker-sat.sh
haasken-hpe, line 1: The changes to this file also should have been made in main already. A rebase on main should fix this.

sat/cli/bootsys/state_recorder.py
```diff
@@ -321,20 +321,24 @@ def do_state_capture(args):
...
-                LOGGER.info(f'Capturing {sr.description}')
-                sr.dump_state()
-            except StateError as err:
-                LOGGER.error(f'Failed to capture {sr.description}: {err}')
-                failed.append(sr)
-
-    if failed:
+    try:
+        state_recorders = [PodStateRecorder()]
+    except Exception as err:
```
haasken-hpe, line 326: It's probably best to catch the specific type of Exception we are looking for, in this case, `S3ResourceCreationError`.

```suggestion
    except S3ResourceCreationError as err:
```

tests/test_util.py
```diff
@@ -659,18 +659,30 @@ def test_get_s3_resource(self):
...
+        try:
+            util.get_s3_resource()
+        except util.S3ResourceCreationError as e:
+            # Check if the exception message contains the expected error message
+            if 'Failed to open file' in str(e):
+                # The exception message matches the expected error message
+                pass
+            else:
+                # The exception message does not match the expected error message
+                raise AssertionError(f"Unexpected exception message: {str(e)}")
```
haasken-hpe, line 671: It's best to use `self.assertRaises` or `self.assertRaisesRegex` like this:

```suggestion
        expected_err = 'Unable to load configuration: Failed to open file'
        with self.assertRaisesRegex(util.S3ResourceCreationError, expected_err):
            util.get_s3_resource()
```

This way, it ensures an exception is raised as expected, and it ensures it has the expected error message. The current code you have does not actually ensure an exception is raised at all. For example, if you comment out line 661 above, it still passes. When using `self.assertRaises` or `self.assertRaisesRegex`, that is not the case, and it will fail if line 661 is removed.

tests/test_util.py
```diff
@@ -659,18 +659,30 @@ def test_get_s3_resource(self):
...
+        try:
+            util.get_s3_resource()
+        except util.S3ResourceCreationError as e:
+            # Check if the exception message contains the expected error message
+            if 'Bad URL value' in str(e):
+                # The exception message matches the expected error message
+                pass
+            else:
+                # The exception message does not match the expected error message
+                raise AssertionError(f"Unexpected exception message: {str(e)}")
```
haasken-hpe, line 685: Same suggestion here:

```suggestion
        expected_err = 'Unable to load S3 API: Bad URL value'
        with self.assertRaisesRegex(util.S3ResourceCreationError, expected_err):
            util.get_s3_resource()
```

sat/cli/setrev/main.py
haasken-hpe, line 1: We should handle the exception raised by `get_s3_resource` in `write_site_data` on line 113 as well. In that case, I think we can just log an error like it does in other cases. Ideally, we'd also be exiting the `sat setrev` call with a non-zero exit status, but you don't need to worry about fixing that here as it's a pre-existing bug.

haasken-hpe (changes requested):
This is looking very close to done.

Once you address my last comments, I'd suggest that you test at least the following commands, and these should be tested when both s3_access_key and s3_secret_key are properly set up and when they are not properly set up:

* `sat showrev`
* `sat setrev`
* `sat bootsys shutdown --stage capture-state`

When you test that, please share the output, so we can verify the behavior together. Be careful not to share any sensitive data (e.g. the contents of s3 access/secret keys) on GitHub or elsewhere.

CHANGELOG.md
```diff
@@ -25,6 +25,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.25.4] - 2023-09-01
+
+### Fixed
+- error handle as warning and proceed
```
haasken-hpe, line 31: This is not specific enough to describe this change to the end user. I'd suggest something like this:

```suggestion
- Update `sat showrev` to log a warning and proceed to display other system
  revision information when S3 credentials are not configured.
- Update `sat setrev` and `sat bootsys` to handle missing S3 credentials more
  gracefully.
```

haasken-hpe (approved):
I'm going to approve this, so that you can merge it when you've resolved the last two comments. Perhaps @Sai-Anirudh-B can review as well and approve in my absence.

I would appreciate if you could squash your commits into a single commit with a good description when you are done resolving comments too.

CHANGELOG.md
shivaprasad-metimath: addressed

shivaprasad-metimath:
Testing on MUG post all comments addressed:

### Without access_key and secret_key:

**1. sat setrev**
image
image
went ahead post throwing a warning, took the inputs too. At the end failed with error as s3 requirements are missing 

**2. sat showrev**
image
Failed with error post warning as setrev is incomplete

**3. sat bootsys**
image
As access key is missing it exited with error

shivaprasad-metimath:
### With access_key and secret_key:

**1.sat setrev**
image
image
didnt throw warning and went ahead with taking inputs 

**2. sat showrev**
image
gave the expected output

**3. sat bootsys**
image
was able to capture the state

haasken-hpe (approved):
Looks good. Feel free to merge!

PR #187, opened 2024-01-31
CRAYSAT-1798: Preserve all port policies in `sat swap`

haasken-hpe:
## Summary and Scope

* CRAYSAT-1798: Preserve all port policies in `sat swap`

  When enabling or disabling a switch or a cable in `sat swap switch` or
  `sat swap cable`, properly handle when a port has multiple port policies
  applied to it. This fixes the issue where only the first port policy
  associated with a port is kept through the disable/enable cycle.

  The strategy for keeping a record of the port policies applied to each
  port is unchanged. That is, for the disable action, for each port
  policy associated with the ports to be disabled, append the
  "sat-offline-" prefix to the policy name, create a new policy
  with the prefixed name which has `"state": "OFFLINE"`, and then apply
  the policy to the port being disabled.

  Note that for ports with more than one policy associated with them (e.g.
  fabric ports which have a fabric policy and a qos policy), this will
  result in multiple policies with the same `"state": "OFFLINE"` data
  being assigned to those ports. According to the Slingshot team, this is
  acceptable and should not cause issues.

  Fix and improve the unit tests for the `get_switch_port_data_list`
  method of the `PortManager`.  This now tests the method for a switch
  with just edge ports, a switch with just fabric ports, and a switch with
  both fabric and edge ports. It also does more consistent mocking by
  mocking the `get_port` method instead of directly mocking the
  `FabricControllerClient` class.

  Fix unit tests for `PortManager.get_jack_port_data_list`. Fix the
  `swap_component` unit tests for the `CableSwapper` and `SwitchSwapper`.

  Fix minor issues in tests including PyCharm code inspection type
  warnings and faulty `assertLogs` usages.

  Test Description:
  The following will be tested on baldar:

  * `sat swap switch --action disable` on an enabled switch
  * `sat swap switch --action disable` on an disabled switch
  * `sat swap switch --action enable` on an disabled switch
  * `sat swap switch --action enable` on an enabled switch

  The same will also be tested for `sat swap cable`.

* CRAYSAT-1798: Remove unused `disruptive` argument from `swap_component`

  The `disruptive` argument was not used in the `swap_component` method of
  the `Swapper` class. The value of the `disruptive` command-line option
  is checked in the function `check_arguments` in `sat.cli.swap.main`
  before calling the appropriate function for the cable, switch, or blade.

  Test Description:
  Unit tests pass. PyCharm is happier.

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.27.3] - 2024-01-31
+
+### Fixed
+- Fixed `sat swap switch` and `sat swap cable` so that they preserve all
+  existing port policies applied to the ports on a switch or a cable across the
+  disable and enable actions. The old behavior was that only the first policy
+  would be preserved, which was a problem for ports with multiple policies
+  configured.
+
 ## [3.27.2] - 2024-01-16
 
 ### Fixed
--- a/sat/cli/swap/cable.py
+++ b/sat/cli/swap/cable.py
@@ -1,7 +1,7 @@
 #
 # MIT License
 #
-# (C) Copyright 2020-2021 Hewlett Packard Enterprise Development LP
+# (C) Copyright 2020-2021, 2024 Hewlett Packard Enterprise Development LP
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
 # copy of this software and associated documentation files (the "Software"),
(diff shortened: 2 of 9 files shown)


haasken-hpe:
I'm opening this for review, but I still have some testing to perform for `sat swap cable`, and I have to update the changelog still as well.

haasken-hpe:
I have finished the testing of `sat swap cable` on the same system where I tested `sat swap switch`, and I have updated the CHANGELOG and declared a new version. This is ready for review and then merge.

shivaprasad-metimath (approved):
Looks fine, should be good to merge

PR #215, opened 2024-04-25
CRAYSAT-1193:Improve create and delete fas snapshot

annapoorna-s-alt:
IM: CRAYSAT-1193
Reviewer: Ryan

Improved snapshot creation by removing expiration time 
Add delete option to prompt user to delete the snapshot if it is no longer needed.

## Summary and Scope

- Polling the snapshot in `sat firmware` resulted in HTTP errors in large clusters.
  Hence, adding the retry option when it consecutively fails for up to 5 times.
- Remove expiration time and add `--delete-snapshot` option to prompt the user to delete the snapshot
  if it is no longer needed. By default, log a message referring the user how to delete the snapshot.

## Issues and Related PRs

_Resolves [CRAYSAT-1193](https://jira-pro.it.hpe.com:8443/browse/CRAYSAT-1193)._

## Testing

### Tested on:

  Baldar

### Test description:

_Test by creating a new snapshot using `sat firmware` command_
_Test by deleting the snapshot using `sat firmware --delete-snapshot <snapshote-name>`

## Risks and Mitigations

_Minimal_

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated

### DIFF
--- a/sat/apiclient/fas.py
+++ b/sat/apiclient/fas.py
@@ -1,7 +1,7 @@
 #
 # MIT License
 #
-# (C) Copyright 2020-2023 Hewlett Packard Enterprise Development LP
+# (C) Copyright 2020-2024 Hewlett Packard Enterprise Development LP
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
 # copy of this software and associated documentation files (the "Software"),
@@ -37,17 +37,6 @@ from sat.xname import XName
 
 inf = inflect.engine()
 LOGGER = logging.getLogger(__name__)
-
-
-def _now_and_later(minutes):
-    """Return now and now + minutes as a pair
-
-    Returns:
-        now, later: datetime pair.
-    """
-    now = datetime.utcnow().replace(microsecond=0)
-    later = now + timedelta(minutes=minutes)
-    return now, later
 
 
 class FASClient(APIGatewayClient):
@@ -300,8 +289,23 @@ class FASClient(APIGatewayClient):
 
         return {name: description for name, description in descriptions.items()}
 
-    def _create_temporary_snapshot(self, xnames=None):
-        """Create a temporary snapshot and return its data.
+    def delete_snapshot(self, snapshot):
+        """Delete a snapshot using the FAS API.
+
+        Args:
+            snapshot (str): The name of the snapshot to be deleted.
+
+        Raises:
+            APIError: If there is an error while attempting to delete the snapshot.
+        """
+        try:
+            self.delete(f'snapshots/{snapshot}')
+
+        except APIError:
+            raise APIError(FASClient.err_no_firmware_found)
+
+    def _create_snapshot(self, xnames=None):
+        """Create a snapshot and return its data.
 
         Args:
             xnames: A list of xnames for which to take the snapshot.
@@ -314,18 +318,17 @@ class FASClient(APIGatewayClient):
         Raises:
             APIError: if FAS returned an APIError when creating the snapshot.
                 This can also raise if:
-                    - The payload from creating the temporary snapshot was
+                    - The payload from creating the snapshot was
                       missing either a 'ready' field while polling, or a
                       'devices' field.
                     - The FAS did not indicate if the snapshot was ready.
         """
-        now, later = _now_and_later(10)
+        now = datetime.utcnow().replace(microsecond=0)
         name = 'SAT-{}-{}-{}-{}-{}-{}'.format(
             now.year, now.month, now.day, now.hour, now.minute, now.second
         )
 
-        # create a system snapshot set to expire in 10 minutes
-        payload = {'name': name, 'expirationTime': later.isoformat() + 'Z'}
+        payload = {'name': name}
 
         # filter on xnames if provided.
         if xnames:
@@ -339,13 +342,22 @@ class FASClient(APIGatewayClient):
         # retrieve it and poll until ready
         response = None
         ready = False
+        consecutive_failures = 0
         while not ready:
-            time.sleep(1)
+            time.sleep(5)
 
             try:
                 response = self.get('snapshots', name).json()
+                consecutive_failures = 0  # Reset consecutive failures on successful response
             except APIError as err:
-                raise APIError('Error when polling the snapshot for "ready" status: {}'.format(err))
+                consecutive_failures += 1
+                if consecutive_failures == 5:  # Maximum consecutive failures
+                    raise APIError('Error when polling the snapshot {}: Exceeded maximum consecutive failures',
+                                   format(err))
+                else:
+                    LOGGER.debug('Error when polling the snapshot {}: Retrying...'.format(err))
+                    continue
+
             except ValueError as err:
                 raise APIError('The JSON received was invalid: {}'.format(err))
 
@@ -375,7 +387,7 @@ class FASClient(APIGatewayClient):
                 was malformed.
         """
         # This can raise APIError
-        response = self._create_temporary_snapshot(xnames)
+        response = self._create_snapshot(xnames)
         try:
             devices = response['devices']
             missing_xnames = set(xnames or []) - set(device.get('xname') for device in devices)
--- a/sat/cli/firmware/main.py
+++ b/sat/cli/firmware/main.py
@@ -1,7 +1,7 @@
 #
 # MIT License
 #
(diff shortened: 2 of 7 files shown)


annapoorna-s-alt:
Please find the detailed output here: 
[CRAYSAT-1193.txt](https://github.com/Cray-HPE/sat/files/15108529/CRAYSAT-1193.txt)

sat/apiclient/fas.py
```diff
@@ -339,13 +353,21 @@ def _create_temporary_snapshot(self, xnames=None):
         # retrieve it and poll until ready
         response = None
         ready = False
-        while not ready:
-            time.sleep(1)
+        consecutive_failures = 0
+        while not ready and consecutive_failures < 5:
```
haasken-hpe, line 357: It looks like the check on `consecutive_failures` is unnecessary because it's handled any time it's incremented on line 365 below.

sat/apiclient/fas.py
```diff
@@ -354,6 +376,8 @@ def _create_temporary_snapshot(self, xnames=None):
             except KeyError:
                 raise APIError('Payload returned from GET to snapshots/name did not have "ready" field.')
 
+        LOGGER.info(f'Use `sat firmware --delete-snapshot` command to delete the snapshot if it is no longer needed')
```
haasken-hpe, line 379: I think you can drop "command" from this message.

```suggestion
        LOGGER.info(f'Use `sat firmware --delete-snapshot` to delete the snapshot if it is no longer needed')
```

sat/cli/firmware/main.py
```diff
@@ -36,6 +36,24 @@
...
 
+def delete_snapshot(client, snapshot):
+    """Delete a snapshot using the FASClient.
+
+    Args:
+        client (FASClient): The FASClient to use.
+        snapshot (str): The name of the snapshot to be deleted.
+
+    Raises:
+            APIError: If there is an error while attempting to delete the snapshot.
```
haasken-hpe, line 47:
```suggestion
        SystemExit: If there is an error while attempting to delete the snapshot.
```

sat/cli/firmware/parser.py
```diff
@@ -53,3 +53,8 @@ def add_firmware_subparser(subparsers):
         '--snapshots', dest='snapshots', metavar='SNAPSHOT', nargs='*',
         help='Describe firmware snapshots. Provide no arguments to this '
              'option to list available snapshots.')
+
+    firmware_parser.add_argument(
```
haasken-hpe, line 57: You might want to look into a way to make this mutually exclusive with the `--snapshots` option.

I don't really like how this command was structured to begin with, namely, I don't like that `sat firmware --snapshots` just lists existing snapshots and `sat firmware --snapshots SNAPSHOT_1 [SNAPSHOT_2 ...]` will show the firmware versions in those snapshots. I think it will be hard to change that behavior in a backwards compatible way.

It would just be nice to avoid any confusion in case somebody thinks that `sat firmware --delete-snapshot` means that it will create a snapshot, show the versions and then delete the snapshot. It might also be confusing that `sat firmware --delete-snapshot SNAPSHOT_1 --snapshots SNAPSHOT_1` will just delete SNAPSHOT_1 and will not show its contents. Perhaps you can think of a better way to structure the command line arguments.

sat/apiclient/fas.py
```diff
@@ -324,8 +339,7 @@ def _create_temporary_snapshot(self, xnames=None):
             now.year, now.month, now.day, now.hour, now.minute, now.second
         )
 
-        # create a system snapshot set to expire in 10 minutes
-        payload = {'name': name, 'expirationTime': later.isoformat() + 'Z'}
+        payload = {'name': name}
```
haasken-hpe, line 342: the `later` variable is no longer used, which might be a little confusing.

One idea might be to rename this method  to `_create_snapshot` and have it take a boolean argument `expires` or `temporary` that controls whether the snapshot expires. Another idea could be to just rename it to `_create_snapshot` and get rid of the expiration logic and the `later` variable entirely.

haasken-hpe (approved):
This looks good, nice job. I have some comments about usability for you to consider and if you want to implement any of those, I will re-review.

An additional usability question I have is whether it's desirable to be able to run `sat firmware --delete-snapshot` and have it just automatically delete the snapshot for you when it's done. Just an idea to consider.

haasken-hpe:
One thing you could try to do to test the resiliency behavior you added would be to start a `sat firmware` command on a decently large system and then between when `sat firmware` issues the POST request to create the snapshot and when it starts polling, perhaps you could force FAS into a bad state, possibly with a rollout restart or something in Kubernetes. The HMS folks could probably help you find a way to test that behavior.

sat/apiclient/fas.py
annapoorna-s-alt: Got it. Removed on line 357.

sat/apiclient/fas.py
annapoorna-s-alt: I prefer to remove the expiration logic completely as it is nowhere used and renaming the function to `create_snapshot`

shivaprasad-metimath (approved):

sat/cli/firmware/parser.py
annapoorna-s-alt: I made `--snapshots` and `--delete-snapshot` as mutually exclusive. So user might not get confused here as they can use only one at a time. So  `sat firmware --delete-snapshot SNAPSHOT_1 --snapshots SNAPSHOT_1` would be invalid.

To be more specific `sat firmware --delete-snapshot SNAPSHOT_1` deletes the snapshot SNAPSHOT_1 if it is valid snapshot-name else throws an error.

annapoorna-s-alt:
### Tested on Odin.
Please find the attached output after addressing the comments.
[CRAYSAT-1193-final.txt](https://github.com/Cray-HPE/sat/files/15148783/CRAYSAT-1193-final.txt)

sat/apiclient/fas.py
```diff
@@ -319,13 +333,12 @@ def _create_temporary_snapshot(self, xnames=None):
                       'devices' field.
                     - The FAS did not indicate if the snapshot was ready.
         """
-        now, later = _now_and_later(10)
+        now = create_snapshot()
```
haasken-hpe, line 336: just remove `create_snapshot` and put the code that gets the date here

```suggestion
        now = datetime.utcnow().replace(microsecond=0)
```

sat/apiclient/fas.py
```diff
@@ -339,13 +352,21 @@ def _create_temporary_snapshot(self, xnames=None):
...
                 response = self.get('snapshots', name).json()
+                consecutive_failures = 0  # Reset consecutive failures on successful response
             except APIError as err:
-                raise APIError('Error when polling the snapshot for "ready" status: {}'.format(err))
+                consecutive_failures += 1
+                if consecutive_failures == 5:  # Maximum consecutive failures
+                    raise APIError('Error when polling the snapshot {}: Exceeded maximum consecutive failures',
+                                   format(err))
+                else:
+                    continue
```
haasken-hpe, line 368: I would suggest logging a debug-level message here.

sat/apiclient/fas.py
```diff
@@ -354,6 +375,8 @@ def _create_temporary_snapshot(self, xnames=None):
             except KeyError:
                 raise APIError('Payload returned from GET to snapshots/name did not have "ready" field.')
 
+        LOGGER.info(f'Use `sat firmware --delete-snapshot` to delete the snapshot if it is no longer needed')
```
haasken-hpe, line 378: Suggest moving this log message to occur after the table is output so the admin will see it.

haasken-hpe (approved):

sat/apiclient/fas.py
annapoorna-s-alt: Done

sat/apiclient/fas.py
annapoorna-s-alt: Addressed

annapoorna-s-alt:
### Tested on Odin.

Please find the attached output after adding the additional feature to delete more than one snapshot
[CRAYSAT-1193-firmware.txt](https://github.com/Cray-HPE/sat/files/15187241/CRAYSAT-1193-firmware.txt)

haasken-hpe (approved):
Looks good!

PR #230, opened 2024-06-26
CRAYSAT-1869: Addressing the value rounding bug

shivaprasad-metimath:
IM:CRAYSAT-1869
Reviewer:Ryan

## Summary and Scope
There is a rounding error when SAT reports status on percentage done. It rounds up to 100% when the actual number reported by BOS is something like 99.9475. For anything other than 100%, most people would not care. But 100% means it is all done, so why would SAT still be waiting for it to go over 100%

Hence, Updating the rounding to 2 decimal places instead of no decimal places and then not rounding if the value above 99 to address the bug and showing upto 10 decimal places.

The same case is added part of the unit test for the class.

## Issues and Related PRs

* Resolves [CRAYSAT-1869](https://jira-pro.it.hpe.com:8443/browse/CRAYSAT-1869)

## Testing

### Tested on:

Would need a large system to validate this changes. hence a unittest case has been added to validate the scenario.
### Test description:

Will have to power off and power on the nodes to verify using a BOS session template

## Risks and Mitigations

minimal

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.28.7] - 2024-06-26
+
+### Fixed
+- Improve info messages logged by `sat bootsys` in the `bos-operations` stage to
+  use more precision when displaying percent successful values in order to avoid
+  prematurely reporting that a BOS session is 100% complete when the actual
+  percentage is between 99 and 100%.
+
 ## [3.28.6] - 2024-06-10
 
 ### Fixed
--- a/sat/cli/bootsys/bos.py
+++ b/sat/cli/bootsys/bos.py
@@ -1,7 +1,7 @@
 #
 # MIT License
 #
-# (C) Copyright 2020-2023 Hewlett Packard Enterprise Development LP
+# (C) Copyright 2020-2024 Hewlett Packard Enterprise Development LP
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
 # copy of this software and associated documentation files (the "Software"),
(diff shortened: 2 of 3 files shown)


sat/cli/bootsys/bos.py
```diff
@@ -255,10 +255,16 @@ def has_completed(self):
                 # Only log progress update when components succeed or fail
                 self.pct_successful = float(self.session_status['percent_successful'])
                 self.pct_failed = float(self.session_status['percent_failed'])
-                LOGGER.info(
-                    'Session %s: %.0f%% components succeeded, %.0f%% components failed',
-                    self.bos_session_thread.session_id, self.pct_successful, self.pct_failed
-                )
+                if self.pct_successful > 99.0 and self.pct_successful < 100.0:
+                    LOGGER.info(
+                        'Session %s: %.10f%% components succeeded, %.2f%% components failed',
```
haasken-hpe, line 260: 10 decimal places is probably overkill. According to [the docs](https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting), the default value of precision is 6. In order for a `self.pct_successful` value which is actually below 100 to be rounded up to 100 when rounding to 6 decimal places, the value would have to be greater than or equal to 99.9999995. That would mean only 5 out of a billion components were not yet successful in a BOS session. I don't expect we will ever boot a billion components at once with BOS.

CHANGELOG.md
```diff
@@ -25,6 +25,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.28.6] - 2024-06-26
+
+### Fixed
+- Updating the rounding to 4 decimal places instead of 2 and then not rounding if the 
+  value is above 99 to address the bug
```
haasken-hpe, line 32: This changelog entry is not clear enough because it does not specify the command affected, so it will not be very helpful to somebody just reading the changelog. Here's what I would suggest:

```suggestion
### Fixed
- Improve info messages logged by `sat bootsys` in the `bos-operations` stage to
  use more precision when displaying percent successful values in order to avoid
  prematurely reporting that a BOS session is 100% complete when the actual
  percentage is between 99 and 100%.
```

haasken-hpe (approved):
Looks good overall, but please apply my suggestions.

sat/cli/bootsys/bos.py
shivaprasad-metimath: Yaa that's right Ryan, I had just considered it for an extreme condition and had taken it as 10. so, should be fine to have it considered at 6 itself as maximum decimal place.

annapoorna-s-alt (approved):

PR #265, opened 2024-09-19
CASMTRIAGE-1900: Add jinja render to rootfs_provider_passthrough to support iSCSI

annapoorna-s-alt:
Reviewer: Annapoorna

## Summary and Scope

In order to suport iSCSI in CSM 1.6. jinja rendering of the rootfs_provider_passthrough value is necessary. Hence the functionality to render the value has been added. 
For both UAN & Compute support has been added for the input yaml to be handled.

## Issues and Related PRs

* Resolves [CRAYSAT-1900](https://jira-pro.it.hpe.com:8443/browse/CRAYSAT-1900)

## Testing

### Tested on:

  * `<Gamora>`

### Test description:

The modified input yaml file consisting of both iSCSI and DVS was considered for testing sat bootprep to create the session templates. It successfully was able to create all the type of templates.

## Risks and Mitigations

minimal

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated

### DIFF
--- a/sat/cli/bootprep/input/base.py
+++ b/sat/cli/bootprep/input/base.py
@@ -1,7 +1,7 @@
 #
 # MIT License
 #
-# (C) Copyright 2021-2023 Hewlett Packard Enterprise Development LP
+# (C) Copyright 2021-2024 Hewlett Packard Enterprise Development LP
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
 # copy of this software and associated documentation files (the "Software"),
@@ -241,6 +241,11 @@ class BaseInputItem(Validatable, ABC):
         # The 'name' property is required by the schema for all types of input
         # items that inherit from BaseInputItem.
         return self.data['name']
+
+    @property
+    def boot_set(self):
+        """Return the full boot_sets dictionary."""
+        return self.data.get('bos_parameters', {}).get('boot_sets', {})
 
     def __str__(self):
         # Since the name can be rendered, and when unrendered, it does not need
--- a/sat/cli/bootprep/input/session_template.py
+++ b/sat/cli/bootprep/input/session_template.py
@@ -186,6 +186,12 @@ class InputSessionTemplate(BaseInputItem):
             boot_set_api_data.update(boot_set_data)
             api_data['boot_sets'][boot_set_name] = boot_set_api_data
 
+            # Fetch full boot sets
+            boot_sets = self.boot_set
+            # Render the rootfs_provider_passthrough using Jinja2
+            rendered_rootfs = self.jinja_env.from_string(
+                        boot_sets[boot_set_name]['rootfs_provider_passthrough']).render(self.data)
+            api_data['boot_sets'][boot_set_name]['rootfs_provider_passthrough'] = rendered_rootfs
         return api_data
 
     def get_image_record_by_id(self, ims_image_id):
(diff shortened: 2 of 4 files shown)


shivaprasad-metimath:
test output from gamora: https://gist.github.com/shivaprasad-metimath/6441895ebdab44ecb19fe7bd27e43593

annapoorna-s-alt (commented):
Looks good with the code changes and testing from session template. Approved!

vinanti-p (approved):

cfawcett-hpe (approved):

sat/cli/bootprep/input/base.py
```diff
@@ -242,6 +242,11 @@ def name(self):
         # items that inherit from BaseInputItem.
         return self.data['name']
 
+    @property
+    def boot_set(self):
```
haasken-hpe, line 246: Why was  a `boot_set` property added on the `BaseInputItem`? This parent class applies to CFS configurations and BOS session templates in the input file, so it's not appropriate to add here. There is also already a `boot_sets` property on the `InputSessionTemplate` class where you needed to use it. Can we just use that one?

sat/cli/bootprep/input/session_template.py
```diff
@@ -186,6 +186,12 @@ def get_create_item_data(self):
             boot_set_api_data.update(boot_set_data)
             api_data['boot_sets'][boot_set_name] = boot_set_api_data
 
+            # Fetch full boot sets
+            boot_sets = self.boot_set
+            # Render the rootfs_provider_passthrough using Jinja2
+            rendered_rootfs = self.jinja_env.from_string(
+                        boot_sets[boot_set_name]['rootfs_provider_passthrough']).render(self.data)
```
haasken-hpe, line 193: Why do we pass in `self.data` to this? From what I can tell from the Jira, it is not required to allow it to render fields with arbitrary data from the session template. I think there's a way we can just modify and use the `jinja_rendered` decorator instead.

I think passing in `self.data` as the context when rendering is going to lead to some weird behavior and possibly hard-to-fix bugs, and then it will be hard to remove the ability to use the data if people start using it that way. I would suggest we not add this unless it's explicitly requested.

haasken-hpe, line 193: Exceptions also aren't caught here which will lead to tracebacks (e.g. if a variable used in a Jinja expression is not set).

tests/cli/bootprep/input/test_session_template.py
```diff
@@ -92,10 +92,12 @@ def get_input_and_expected_bos_data(self, name='my-session-template',
         bos_payload_bootsets = {}
         for boot_set_name, boot_set_arch in arch_by_bootset.items():
             input_bootsets[boot_set_name] = {
-                'node_roles_groups': ['Compute']
+                'node_roles_groups': ['Compute'],
+                'rootfs_provider_passthrough': "dvs:api-gw-service-nmn.local:300:hsn0,nmn0:0"
             }
             bos_payload_bootsets[boot_set_name] = {
                 'node_roles_groups': ['Compute'],
+                'rootfs_provider_passthrough': "dvs:api-gw-service-nmn.local:300:hsn0,nmn0:0",
```
haasken-hpe, line 100: It would be good to add a test where the rootfs_provider_passthrough has a variable that needs to be rendered with Jinja2

haasken-hpe (commented):
I think it would be good to fix my comments. I opened CRAYSAT-1917 and pushed a branch with a possible fix.

PR #288, opened 2024-11-15
CRAYSAT-1841: Add support for CFS v2 or v3 to `sat bootprep`

haasken-hpe:
## Summary and Scope

Add support for using either the CFS v2 or v3 API in `sat bootprep` based on the value of the command-line option `--cfs-version` or the config-file option `cfs.api_version`. Before this change, `sat bootprep` was hard-coded to always use CFS v2.

Update the `InputConfiguration`, `InputConfigurationLayer`, and `AdditionalInventory` classes to keep a reference to the instance of the version-specific subclass of `CFSClientBase`, so that they can delegate operations to the appropriate version of the CFS API.

This removes a lot of code that required knowledge of the CFS API and instead uses classes and methods defined in the `csm_api_client` library, which has already been updated to support both CFS v2 and v3. This should reduce code duplication across `sat` and `csm_api_client`.

Some unit tests are no longer needed here because that functionality is tested in the `csm_api_client` instead. Other unit tests are updated to include the `cfs_client` instance that needs to be passed into the input layer and configuration objects.

This also makes the `playbook` property required in the schema because the CFS v3 API requires that the playbook be specified. This is technically a backwards-incompatible schema change because it can invalidate certain bootprep input files if they define layers without playbooks. The default bootprep files contain a playbook for every layer and do not depend on the CFS default playbook value, so this should be a pretty safe and non-disruptive change.

## Issues and Related PRs

* Resolves CRAYSAT-1896

### Tested on:

### Test description:

Unit tests only so far.

Still needs to be tested on a real system.

## Risks and Mitigations

This has a moderate risk level. It introduces a bootprep input file schema change that can cause files that do not specify playbook to not validate anymore.

It also is a significant code refactor in that it pushes a lot of the CFS-API knowledge out of the bootprep code and offloads it to the `csm_api_client` library.

These risks can be mitigated by thorough testing to ensure that CFS configuration creation and image customization still works in `sat bootprep`.

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated


haasken-hpe:
Keeping this as a draft until I test, but it's a fairly large PR, so feel free to look at it before I finish testing and let me know if you have any questions or concerns.

haasken-hpe:
It looks like I need to still add support for pagination of resources other than sessions (e.g. configurations). We do a GET of `/configurations` when we are checking whether any of the configurations that are going to be created by the input file already exist in CFS.

So I need to add paging support for those other endpoints to `python-csm-api-client` and then update this code to use that. It should be a pretty straightforward change. I'll work on that next week.

haasken-hpe:
I also still need to add the `--cfs-version` command-line option to the `bootprep` command.

haasken-hpe:
I appear to have broken the ability to specify "latest" as the product version in a configuration layer as well. I'll have to look into that.

haasken-hpe:
I have fixed the issues noted in my prior comments. That is:

* Fixed the ability to specify the "latest" value as the product version in a product-based config layer
* Added the missing `--cfs-version` option to `sat bootprep run`
* Fixed the listing of CFS configurations with the CFS v3 API to handle paging

haasken-hpe:
Here are my testing results from rocket: https://gist.github.com/haasken-hpe/8cc8455889b7afe729c5b77dcebfc0f7

I have only done limited testing with `--cfs-version v2`, and I'm just using the v3 default in these tests, but I have done some minimal testing with v2, and I think we can be sure to test this with our automated functional tests we are going to write.

annapoorna-s-alt (approved):

haasken-hpe:
Since I have merged https://github.com/Cray-HPE/python-csm-api-client/pull/47 and have pushed a release/2.3 branch there, we now have a stable build, so I am rebasing this PR to drop the commit that references the unstable repo.

ethanholen-hpe (approved):

PR #295, opened 2024-12-02
CRAYSAT-1936: add ability to sort reports by multiple fields

ethanholen-hpe:
## Summary and Scope

Added ability to sort reports by multiple fields

## Issues and Related PRs

* Resolves [CRAYSAT-1936](https://jira-pro.it.hpe.com:8443/browse/CRAYSAT-1936)

### Tested on:

  * drax

### Test description:

_How were the changes tested and success verified? If schema changes were part of this change, how were those handled in your upgrade/downgrade testing?_

- Were continuous integration tests run? If not, why? yes
- Was upgrade tested? If not, why? yes
- Was downgrade tested? If not, why? yes
- Were new tests (or test issues/Jiras) created for this change? yes

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated

### DIFF
--- a/sat/report.py
+++ b/sat/report.py
@@ -1,7 +1,7 @@
 #
 # MIT License
 #
-# (C) Copyright 2019-2023 Hewlett Packard Enterprise Development LP
+# (C) Copyright 2019-2024 Hewlett Packard Enterprise Development LP
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
 # copy of this software and associated documentation files (the "Software"),
@@ -93,9 +93,8 @@ class Report:
         Args:
             headings: Headings for the table's columns.
             title: Title for the table
-            sort_by: Sort the output by the desired column when printing
-                in tabular format. Can be the name of a heading, or a 0-based
-                index.
+            sort_by: List of columns to sort by when printing in tabular format.
+                Each column can be the name of a heading, or a 0-based index.
             reverse: If True, then reverse the sorting order.
             no_headings: If True, then omit the title block and column
                 headings from the display.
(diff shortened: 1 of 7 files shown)


ethanholen-hpe:
Output from testing on drax: https://gist.github.com/ethanholen-hpe/a4bc6a67dc7076ebb164c8b8b14644a4

sat/report.py
```diff
@@ -159,19 +158,23 @@ def __init__(self, headings, title=None,
...
-                self.sort_by = match_query_key(self.sort_by, headings)
-                if not self.sort_by:
+            for i in range(len(self.sort_by)):
+                try:
+                    self.sort_by[i] = int(self.sort_by[i])
+                    self.sort_by[i] = self.headings[self.sort_by[i]]
+                except IndexError:
+                    # sort_by is out of range.
                     LOGGER.warning(warn_str, sort_by, self.headings)
+                    self.sort_by = None
```
haasken-hpe, line 171: We might want to consider ignoring only the `sort_by` elements that are not valid, rather than ignoring all of them. In that case, you might need to generate a new, converted list in this for loop. For invalid elements of `sort_by`, you would just skip adding them to that new list with a logged warning. Then at the end of the loop, you'd set `self.sort_by` to this new list. If the new list is empty, then it means none of the given `sort_by` values were valid, so then you'd set `self.sort_by = None` and log the warning that the output will be unsorted.

This is a pretty minor edge case though. I could see it maybe being useful for the commands we have that actually print multiple reports (e.g. `sat showrev`) where those reports have different headings. E.g., the `--products` output vs. the `--local` output have different headings, and a comma-separated `--sort-by` value could be used to sort the two different reports by different headings.

tests/test_report.py
```diff
@@ -390,6 +390,17 @@ def test_sorted_print(self):
         for expected, actual in zip(self.entries, pt_s):
             self.assertEqual(expected, actual)
 
+    def test_sorted_list_print(self):
+        """The internal PT should be sorted on the first column.
```
haasken-hpe, line 394:
```suggestion
        """The internal PT should be sorted on the first column and then the second column
```

tests/test_report.py
```diff
@@ -390,6 +390,17 @@ def test_sorted_print(self):
...
+        """
+        report = Report(self.headings, sort_by=[0, 1])
+
+        report.add_rows(self.out_of_order)
+        pt_s = get_report_printed_list(report)
+
+        for expected, actual in zip(self.entries, pt_s):
+            self.assertEqual(expected, actual)
+
```
haasken-hpe, line 403: The test you added is a good one to show that asking it to sort with `0,1` sorts first on column 1 putting it in the order `[e3, e1, e2]`, and then sorts on column 0, which puts it in the order `[e1, e2, e3]`, but it doesn't test that it does a stable sort. This test would check that minimally:

```suggestion

    def test_sorted_list_with_matches_print(self):
        """Test sort on first and second column when elements are equal in first column
        """
        e1 = ['alice', 'mars', 'red']
        e2 = ['bob', 'venus', 'blue']
        e3 = ['charlie', 'earth', 'purple']
        e4 = ['alice', 'earth', 'green']
        out_of_order = [e1, e3, e2, e4]
        expected_order = [e4, e1, e2, e3]
        report = Report(self.headings, sort_by=[0, 1])
        report.add_rows(out_of_order)
        
        pt_s = get_report_printed_list(report)

        self.assertEqual(expected_order, pt_s)
```

haasken-hpe (approved):
Looks good. My comments are minor.

I would suggest also testing behavior when given one or more unrecognized fields for `--sort-by`. This relates to one of my comments on how we handle that as well.

ethanholen-hpe:
Issue happening when running sat showrev: https://gist.github.com/ethanholen-hpe/445b90d6cc9349c52a764b510b0be5ae

ethanholen-hpe:
Testing and fix for issue: https://gist.github.com/ethanholen-hpe/c2017cb8402048ad9ff4a43dfb16e9fd

